PR #9, opened 2015-05-14
pt-archiver corrected excessively frequent lag check

frank-cizmich:
pt-archiver was too aggressive checking for slave lag, doing it for every row inserted.
Now it checks every 100 rows. Speed improved substantially.

### DIFF
--- a/bin/pt-archiver
+++ b/bin/pt-archiver
@@ -6074,6 +6074,7 @@ sub main {
    # This row is the first row fetched from each 'chunk'.
    my $first_row = [ @$row ];
    my $csv_row;
+   my $lag_count = 0;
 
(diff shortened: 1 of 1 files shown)


daniel-nichter:
I'd mention this in the docs. Any way to test this? Otherwise, :+1:

frank-cizmich:
Ack. Mentioned in doc, but in another branch that already addresses  --check-interval doc issue.
See:  https://github.com/percona/percona-toolkit/pull/6/files

Testing this seems like 100 times more complicated than the code fix itself.
Maybe setting a --limit = 1 , setting an artificial lag (available on 5.6) .. and measuring time. 
Ugh!

PR #10, opened 2015-05-15
pt-osc fixed ReadKeyMini issue

frank-cizmich:
Removed ReadKeyMini that caused trouble in some circumstances:
https://bugs.launchpad.net/percona-toolkit/+bug/1450499

Originally included to solve inconsistent terminal state when tool was interrupted during the ask password phase.
https://bugs.launchpad.net/percona-toolkit/+bug/1396870

We now use Term::ReadKey for this. Since this module's availability is not guaranteed, we include it within an eval, and use it's ReadMode function conditional to it having been successfully imported.

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -40,7 +40,6 @@ BEGIN {
       HTTP::Micro
       VersionCheck
       Percona::XtraDB::Cluster
-      ReadKeyMini
    ));
 }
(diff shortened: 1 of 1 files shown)


daniel-nichter:
:+1: If you say it works, good enough for me. :-) `ReadKeyMini` was a custom solution that I never understood well.

PR #13, opened 2015-05-21
pt-sift failed to download missing tools

frank-cizmich:
pt-sift uses some other tools so it checks if they are present and downloads them using "curl" if missing.
This started to fail once percona.com website protocol changed to https.
Solved changing to http->https and also adding redirect option to curl "-L" for good measure.


daniel-nichter:
:+1:

PR #18, opened 2015-06-12
Pt stalk mongodb

fipar:
This adds very bare bones collection for mongodb/tokumx (I've tested it successfully against both). 

It should be improved by capturing more things (and perhaps deciding what to capture depending on the connection type, as, for example, it currently will attempt to collect sh.status() and rs.status() in all cases), but this may be a good usable/working starting point. 

It is missing tests, as that would require setting up a mongodb/tokumx sandbox, but it is not breaking any of the existing pt-stalk tests. 

It adds the --mongo option which, if provided, allows --function to be only one of: 
- a file name (so that we can implement any kind of mongo checks via plugins), or
- nothing, in which case a mongodb specific default is used (trg_mongo_defaults, which checks the provided --threshold against the number of active operations as reported by db.currentOp()). 

Also, if --mongo is found, all mysql-related captures are disabled.


svetasmirnova:
This is same as https://github.com/percona/percona-toolkit/pull/163, so closing this as duplicate.

PR #20, opened 2015-06-19
enable support for RDS in pt-kill

SingerWang:
The standard way to kill a query in MySQL is to run the command "kill query [thread-id]" and the way to kill a connection is the command "kill [thread-id]". The script pt-kill uses the earlier if you specify --kill-query and the later if you specify --kill in the options.

Unfortunately due to limited privileges, we are not able to run these commands in RDS. RDS provides stored procedure analogous to those two commands for us to use. These procedures run using a system similar to setuid to run the commands as a more privileged user. The procedure for "kill [thread-id]" is "CALL mysql.rds_kill( [thread-id] )" and "CALL mysql.rds_kill_query( [thread-id] )" for "kill query [thread-id]".

http://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Appendix.MySQL.CommonDBATasks.html

This patch enables RDS support for pt-kill. It adds a new flag ```--rds``` which makes the pt-kill script use the RDS procedure calls instead.

### DIFF
--- a/bin/pt-kill
+++ b/bin/pt-kill
@@ -7642,6 +7642,13 @@ Note that this is a digest (or hash) of the query's "fingerprint",
 so queries of the same form but with different values will have the same ID.
 See pt-query-digest for more information.
 
+=item --rds
+Denotes the instance in question is on Amazon RDS. By default pt-kill runs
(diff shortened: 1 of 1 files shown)


frank-cizmich:
Looks reasonable enough. :-)

SingerWang:
thank you!

PR #23, opened 2015-07-26
pt-archiver multple slave lag and OptionParser repeatable option

frank-cizmich:
Added the the possibility to mark an option as "repeatable" in OptionParser module.
This allows for options to be specified multiple time to create an array, instead of overwriting the previous definition.
This was then used in pt-archiver to be able to define multiple --check-slave-lag options and to check those slaves in a loop. (separating multiple DSNs by a comma was not practical due to regex limitations and the DSN format)
Note that this does not affect previous behaviour of tools because only options marked as "repeatable" are affected, and this was used only in pt-archiver for the --check-slave-lag option, and handled specially for it.
The ability to mark an option as repeatable may be used later on for other options (e.g --set-vars)

### DIFF
--- a/bin/pt-archiver
+++ b/bin/pt-archiver
@@ -784,6 +784,7 @@ sub new {
       'default'    => 1,
       'cumulative' => 1,
       'negatable'  => 1,
+      'repeatable' => 1,  # means it can be specified more than once
    );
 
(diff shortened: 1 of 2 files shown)


daniel-nichter:
:+1:

PR #24, opened 2015-07-30
pt-osc now limits size of contraint names

frank-cizmich:
Problem: 
Every time pt-online-schema runs and --alter-foreign-keys-method=rebuild_constraints , the names of the contraints in the new table are prefixed with an underscore (to avoid conflict).
This is done for both the constraints in the altered table and for the constraints in the child tables.
If the tool is run routinely, the name of the constraints can get very long and even exceed the allowed limit.

Solution: 
Change the renaming policy. If the name has no prefixed underscore, add one. If the name already has a prefixed underscore, remove it.

A test was added to test this new behavior
 
https://bugs.launchpad.net/percona-toolkit/+bug/1215587

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -9861,7 +9861,17 @@ sub create_new_table {
       my $quoted = $q->quote($orig_tbl->{db}, $new_table_name);
       my $sql    = $ddl;
       $sql       =~ s/\ACREATE TABLE .*?\($/CREATE TABLE $quoted (/m;
-      $sql       =~ s/^  CONSTRAINT `/  CONSTRAINT `_/gm;
+
+
+      # If it has a leading underscore, we remove one, otherwise we add one
+      # This is in contrast to previous behavior were we added underscores
+      # indefinitely, sometimes exceeding the allowed name limit
+      # https://bugs.launchpad.net/percona-toolkit/+bug/1215587 
+      if ( $sql =~ /CONSTRAINT `_/ ) {
+         $sql =~ s/^  CONSTRAINT `_/  CONSTRAINT `/gm;
+      } else {
+         $sql =~ s/^  CONSTRAINT `/  CONSTRAINT `_/gm;
+      }
       if ( $o->get('default-engine') ) {
          $sql =~ s/\s+ENGINE=\S+//;
       }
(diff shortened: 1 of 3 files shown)


bin/pt-online-schema-change
```diff
@@ -8562,7 +8562,7 @@ sub main {
          #   at /Users/daniel/p/pt-osc-2.1.1/lib/PerconaTest.pm line 559.
          #                   ''
          #   doesn't match '(?-xism:Failed to find a unique new table name)'
-         my $original_error = $EVAL_ERROR;
+         #my $original_error = $EVAL_ERROR;
```
daniel-nichter: Dead code?

bin/pt-online-schema-change
```diff
@@ -8571,7 +8571,7 @@ sub main {
                warn "Error cleaning up: $EVAL_ERROR\n";
             }
          }
-         die $original_error if $original_error;  # rethrow original error
+         #die $original_error if $original_error;  # rethrow original error
```
daniel-nichter: Dead code?

bin/pt-online-schema-change
```diff
@@ -9861,7 +9861,15 @@ sub create_new_table {
       my $quoted = $q->quote($orig_tbl->{db}, $new_table_name);
       my $sql    = $ddl;
       $sql       =~ s/\ACREATE TABLE .*?\($/CREATE TABLE $quoted (/m;
-      $sql       =~ s/^  CONSTRAINT `/  CONSTRAINT `_/gm;
+
+
+      # If it has a leading underscore, we remove one, otherwise we add one
```
daniel-nichter: Why? I.e. comment doesn't explain why we flip-flop the `_`.

daniel-nichter:
I'm not sure about this approach. FK constraints are identifiers. Granted, they're probably not widely used, but nonetheless we're re-using identifiers and I wonder if it might somehow cause problems because now there's only 2 identifiers that can regularly alternate. The successive addition of `_` doesn't have this risk. A counter inside the FK constraint might be better, e.g. `_N_ID` and just keep doing `N++`.

When new logic like this is introduced, full integration tests are good, but factoring out the logic and unit testing it is a very good practice. For example, if this logic was factored out, and presuming the interface well-defined, we could implement any log (successive `_`, flip-flop `_`, or `_N++`) without touching the rest of the code.

I leave these considerations/decisions ^ to you. In general, this should be safe enough, and backwards-compatibility isn't an issue here.

bin/pt-online-schema-change
frank-cizmich: Yes. It actually is a fix for another bug that made debugging this fix difficult. Will remove it as it will get merged anyway.

bin/pt-online-schema-change
frank-cizmich: Noted.

frank-cizmich:
I'm not sure that flip-flopping is actually worse than tagging '_'s indefinitely. I don't see the chance of conflict increasing much.
The ultimate solution would be to tag a random string, as is mentioned in a comment in the code. "this code isn't perfect ... should we generate random/UUID FK names or something instead?"
About putting the logic in one place, it sounds great but looking at the code, it isn't entirely trivial.
I'll keep as is for now.

PR #34, opened 2015-08-26
pt-slave-restart fix for MariaDB 10.0 and above

tanji:
pt-slave-restart is broken on MariaDB 10.0 and above, because VersionParser only checks if the version number is superior to 5.6.5. By checking as well that version number is also inferior to 10.0.0, that allows pt-slave-restart to function with MariaDB 10.0 and above, given that GTID replication errors can be skipped safely using the Sql_slave_skip_counter method.

### DIFF
--- a/bin/pt-slave-restart
+++ b/bin/pt-slave-restart
@@ -5004,7 +5004,7 @@ sub watch_server {
    # transaction is not possible with SQL_SLAVE_SKIP_COUNTER
    my $skip_event;
    my $have_gtid = 0;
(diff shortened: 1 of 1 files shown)


dveeden:
Looks identical to my fix ( #42  ). 
Any reason this isn't merged yet?

frank-cizmich:
Tested and approved, but will use different pull request:
https://github.com/percona/percona-toolkit/pull/62 ... so closing this one
Will be included in next release.
Thanks all.

PR #41, opened 2015-10-21
Fix #1508401 Specify GPL Version in the RPm metadata

roidelapluie:
The License field in the Percona-toolkit RPM is GPL. This does not tell
the version of GPL and might confuse tools and users that rely on this
field.

Best practice in the RPM world is to use "GPLv2" for the GPL 2.0
license.

The commit switches the license field of the RPM to GPLv2.


roidelapluie:
@frank-cizmich any news on this?

frank-cizmich:
Sorry for delay. Will consult with packaging guys and , probably, EMT today

PR #45, opened 2015-10-28
pt-osc handle slave disconnect without exiting (LP 1402051)

daniel-nichter:


daniel-nichter:
Superseded by https://github.com/percona/percona-toolkit/pull/51

PR #48, opened 2015-10-30
Changing default character set of the checksums table to utf8

jynus:
Not setting explicitly the character set breaks the checksums
when using a default binary collation (the standard on all
Wikimedia sites). MySQL 5.5 and above has a system charset
of utf8 and it is not possible to change it without recompiling.

Bug: https://bugs.launchpad.net/percona-toolkit/+bug/1485195

### DIFF
--- a/bin/pt-table-checksum
+++ b/bin/pt-table-checksum
@@ -12491,7 +12491,7 @@ structure (MAGIC_create_replicate):
      ts             TIMESTAMP    NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
      PRIMARY KEY (db, tbl, chunk),
      INDEX ts_db_tbl (ts, db, tbl)
-  ) ENGINE=InnoDB;
+  ) ENGINE=InnoDB DEFAULT CHARSET=utf8;
 
 Note: lower_boundary and upper_boundary data type can be BLOB. See L<"--binary-index">.
 


dveeden:
Why not set this to utf8mb4?

jynus:
mysql system (identifiers) encoding, up to 5.6 is utf8 (I have not checked), utf8mb4 may be ok, but not needed.

Probably it is better to set it on the tbl and db fields, as the other text value may need to be different.

percona-csalguero:
Thanks for the fix @jynus. This fix has been targeted to be released in Percona Toolkit 2.2.19 but before merging it into the 2.2 branch, we need to add a test case.

jynus:
```
mysql -e "DROP DATABASE IF EXISTS my_binary_database; CREATE DATABASE my_binary_database CHARSET=binary"
pt-table-checksum --replicate my_binary_database.my_table
if mysql my_binary_database -e "SHOW CREATE TABLE my_table\G" | grep tbl | grep binary; then echo "The problem still exists"; fi
```

I herby declare my patch and the above 3 lines of bash to be in the Public Domain, if that was not possible, as CC0 license. Feel free to integrate that on the testing framework or use it in any way you want.

percona-csalguero:
Thanks again @jynus.
I'll put a test based on your code above into the utility's test suite (https://github.com/percona/percona-toolkit/tree/2.2/t/pt-table-checksum)

PR #53, opened 2015-11-04
Docs table regex

dveeden:
`--tables-regex='^abc$'` matches abc, Abc, etc. but `--tables-regex='^Abc$'` won't match any of those.
Also using /regex/ vs. regex is not clear.

### DIFF
--- a/bin/pt-table-checksum
+++ b/bin/pt-table-checksum
@@ -12165,7 +12165,9 @@ Only checksum this comma-separated list of databases.
 
 type: string; group: Filter
 
-Only checksum databases whose names match this Perl regex.
+Only checksum databases whose names match this Perl regex. This is matched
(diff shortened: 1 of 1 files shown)


dveeden:
Another solutions would be to make pt-table-checksum
* lowercase the value got from --tables-regex
* Do a full case sensitive comparison so we could differntiate between Abc and abc.
* Allow/detect full perlre (e.g. '/abc/i') if the first and last characters are slashes.

svetasmirnova (approved):

dveeden:
@svetasmirnova the "clahub" check might need some attention. When I click on the "Details" page I get to a website that doesn't seem to have anything to do with the CLA. Maybe the check is old and needs a re-run?

PR #54, opened 2015-11-04
Report about slaves even when no differences are found

dveeden:
If there are no differences `pt-table-checksum --replicate-check-only` doesn't give any output. It might set a correct return code. But I'd like to know which slaves were checked.


svetasmirnova (changes requested):

Could you please add a test case? You can find examples in the directory t.

dveeden:
Let's close this for now. I'm not using this tool at the moment and it looks like nobody else had a need for this.

PR #55, opened 2015-11-09
Release 2.2.16

tplavcic:
This is a release branch for PT 2.2.16.


tplavcic:
I have created this PR from release branch to development branch as we discussed but we should think would it be better to create it to 2.2 branch and then merge 2.2 back to development - this is because some new commits might happen on development branch around release and if we follow release->development->2.2 they might end up in 2.2 branch even if they're not released yet.
release->2.2->development might be better in this case

PR #60, opened 2015-12-14
pt-osc fixed recursion method dsn - lp1523685

frank-cizmich:
pt-osc recursion-method choked on dsn method.
Customized MasterSlave module was the culprit. Apparently It was customized for pt-osc (and pt-query-digest) and when tools were updated customization was overwritten.
Fixed by modifying the module with the enhancement and applying it to pt-osc (and pt-query-digest)
Created unit test for pt-osc for this.


frank-cizmich:
https://bugs.launchpad.net/percona-toolkit/+bug/1523685

PR #62, opened 2015-12-30
pt-slave-restart now checks for MariaDB version

frank-cizmich:
background: pt-slave-restart checks for MySQL version and if gtid is enabled to handle skipping in a different way accordingly.
MariaDB uses a different versioning number system so this failed for that database.
We now simply check for both version numbers.


percona-csalguero:
Fixes https://bugs.launchpad.net/percona-toolkit/+bug/1362942

PR #78, opened 2016-02-18
add 5.7 compatibility for pt-show-grants

frank-cizmich:
5.7 changes a lot of the ways users are created and passwords are stored.
To make pt-show-grants compatible we can add the new "SHOW CREATE USER" command to be able to recreate users and permissions from one database to another.


daniel-nichter:
Might be good to put `compare_versions` into some common module/code and unit test it, otherwise :+1: .

frank-cizmich:
done :-)

PR #90, opened 2016-05-14
fix lp:1581752 5.7 compatibility for pt-query-digest, slow log time format

ihanick:
https://bugs.launchpad.net/percona-toolkit/+bug/1581752

Time field in slow log in 5.7 has new format:
"Time: 2016-05-14T05:28:50.060335Z"

Thus --since and --until pt-query-digest options producing empty result.

### DIFF
--- a/bin/pt-query-digest
+++ b/bin/pt-query-digest
@@ -4993,7 +4993,7 @@ sub new {
    return bless $self, $class;
 }
 
-my $slow_log_ts_line = qr/^# Time: ([0-9: ]{15})/;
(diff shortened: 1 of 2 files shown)


percona-csalguero:
Thanks @ihanick .
I am closing this PR since it was fixed in https://github.com/percona/percona-toolkit/pull/93

percona-csalguero:
Reopening because the fix I've mentioned before only applies to the General Log.

nicholascp:
APPLIES TO SLOW QUERY LOG 
JUST CHANGE  set global log_timestamps='system';
SINCE 1H WILL BE ok

PR #92, opened 2016-05-28
Added --preserve-triggers to pt-osc

ManjotS:
I have added the --preserve-triggers flag to pt-online-schema-change.

With this flag, pt-osc will preserve and edit triggers that already exist on a table. This has been tested on mysql < 5.7 and in sysbench with concurrency.

This gets rid of one shortcoming with pt-osc in that it cannot work in environments with triggers.

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -2253,6 +2253,7 @@ sub get_dbh {
       PrintError         => 0,
       ShowErrorStatement => 1,
       mysql_enable_utf8 => ($cxn_string =~ m/charset=utf8/i ? 1 : 0),
+      mysql_multi_statements => 1,
    };
    @{$defaults}{ keys %$opts } = values %$opts;
    if (delete $defaults->{L}) { # L for LOAD DATA LOCAL INFILE, our own extension
@@ -10607,40 +10608,33 @@ sub create_triggers {
 
       if($delete_trigger_ins_point) {
          $delete_trigger
-         = "LOCK TABLES $orig_tbl->{name};\n"
+         = "LOCK TABLES $orig_tbl->{name} WRITE;\n"
          . "DROP TRIGGER IF EXISTS `$orig_tbl->{db}`.`$orig_delete_trigger->[1]`;\n"
-         . "DELIMITER //\n"
          . $delete_trigger . "\n"
          . substr($orig_delete_trigger->[3], 0, $delete_trigger_ins_point)
          . $delete_trigger_action . ";\n"
-         . substr($orig_delete_trigger->[3], $delete_trigger_ins_point) . "; //\n"
-         . "DELIMITER ;\n"
+         . substr($orig_delete_trigger->[3], $delete_trigger_ins_point) . ";\n"
          . "UNLOCK TABLES;";
       }
       else {
          $delete_trigger
-         = "LOCK TABLES $orig_tbl->{name};\n"
+         = "LOCK TABLES $orig_tbl->{name} WRITE;\n"
          . "DROP TRIGGER IF EXISTS `$orig_tbl->{db}`.`$orig_delete_trigger->[1]`;\n"
-         . "DELIMITER //\n"
          . $delete_trigger . "\n"
          . "BEGIN\n"
          . $orig_delete_trigger->[3] . "\n"
          . $delete_trigger_action . ";\n"
-         . "END; //\n"
-         . "DELIMITER ;\n"
+         . "END;\n"
          . "UNLOCK TABLES;";
       }
-      print $delete_trigger;
+
       $orig_triggers{del}
-         = "LOCK TABLES $orig_tbl->{name};\n"
+         = "LOCK TABLES $orig_tbl->{name} WRITE;\n"
          . "DROP TRIGGER IF EXISTS `${prefix}_del`;\n"
-         . "DELIMITER //\n"
          . "CREATE DEFINER=$orig_delete_trigger->[2] TRIGGER `$orig_delete_trigger->[0]`.`$orig_delete_trigger->[1]` "
          . "AFTER DELETE ON $orig_tbl->{name}\n"
          . "FOR EACH ROW\n"
-         . $orig_delete_trigger->[3] . "\n"
-         . "//\n"
-         . "DELIMITER ;\n"
+         . $orig_delete_trigger->[3] . ";\n"
          . "UNLOCK TABLES;";
    }
    else {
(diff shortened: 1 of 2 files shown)


grypyrg:
Before it can be accepted, I think:
- The code in the individual classes should be modified, not directly in the binary (the binaries are generated)
- You should include tests

ManjotS:
Where are the classes?

grypyrg:
`./lib`

grypyrg:
Some notes after checking the code:

* DSNParser package is changed in `./lib`
* The original trigger does not necessary still work with the new trigger. Hence making this a default feature and not including many many warnings can break things easily.
* `BEGIN END` is something that can be nested in the original triggers already? Does the code handle this? (test-case)
* When do you add the triggers to the new table? Are we doing this atomically? We need to avoid this at all cost. (needs multiple tests to validate it or there might be data inconsistencies because of it)

ManjotS:
* DSNParser package is changed in |./lib|
      o | I will work on this change|
  * The original trigger does not necessary still work with the new
    trigger. Hence making this a default feature and not including many
    many warnings can break things easily.
      o why?
  * |BEGIN END| is something that can be nested in the original triggers
    already? Does the code handle this? (test-case)
      o Yes this is handled and tested
  * When do you add the triggers to the new table? Are we doing this
    atomically? We need to avoid this at all cost. (needs multiple tests
    to validate it or there might be data inconsistencies because of it)
      o New triggers are added same time as always (sub create_triggers)
        except the tables are locked during the change.

grypyrg:
> * DSNParser package is changed in |./lib|
>       o | I will work on this change|
>   * The original trigger does not necessary still work with the new
>     trigger. Hence making this a default feature and not including many
>     many warnings can break things easily.
>       o why?

Triggers work with OLD/NEW rows from their tables. when removing a column, the triggers might break completely. 

> 
>   * |BEGIN END| is something that can be nested in the original triggers
>     already? Does the code handle this? (test-case)
>       o Yes this is handled and tested

`t/` :-)

>   * When do you add the triggers to the new table? Are we doing this
>     atomically? We need to avoid this at all cost. (needs multiple tests
>     to validate it or there might be data inconsistencies because of it)
>       o New triggers are added same time as always (sub create_triggers)
>         except the tables are locked during the change.
> 

Let me try to phrase it better:
When the tables need to be swapped, we need to recreate the triggers on the new table. Does that happen atomically or is there a moment where triggers won't be active?

michaeldg:
The solution currently does not automatically add the triggers. This means the triggers could be executed twice (once on the old table and once on the new table). Perhaps this could be an inspiration to make it dynamically? 
http://code.openark.org/blog/mysql/solving-the-facebook-osc-non-atomic-table-swap-problem 
I think the same principle could be implemented here.

michaeldg:
Also, I have  ran into an issue trying to run the tool:

root@ui-oltp01:~/percona-toolkit# ./bin/pt-online-schema-change -h127.0.0.1 --execute D=db,t=c_s_t --alter="ENgine=InnoDB" --preserve-triggers
Cannot connect to D=db,h=1.2.3.4
No slaves found.  See --recursion-method if host ui-oltp01 has slaves.
Not checking slave lag because no slaves were found and --check-slave-lag was not specified.
Operation, tries, wait:
  analyze_table, 10, 1
  copy_rows, 10, 0.25
  create_triggers, 10, 1
  drop_triggers, 10, 1
  swap_tables, 10, 1
  update_foreign_keys, 10, 1
Altering `db`.`c_s_t`...
Creating new table...
Created new table db._c_s_t_new OK.
Altering new table...
Altered `db`.`_c_s_t_new` OK.
2016-06-08T12:09:37 Creating triggers...
2016-06-08T12:09:37 Dropping triggers...
2016-06-08T12:09:37 Dropped triggers OK.
2016-06-08T12:09:37 Dropping new table...
2016-06-08T12:09:37 Error dropping new table `db`.`_c_s_t_new`: DBD::mysql::db do failed: Table '_c_s_t_new' was locked with a READ lock and can't be updated [for Statement "DROP TABLE IF EXISTS `db`.`_c_s_t_new`;"] at ./bin/pt-online-schema-change line 8909.

To try dropping the new table again, execute:
DROP TABLE IF EXISTS `db`.`_c_s_t_new`;
2016-06-08T12:09:37 Dropped new table OK.
`db`.`c_s_t` was not altered.
Error creating triggers: 2016-06-08T12:09:37 DBD::mysql::db do failed: You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near 'beginIF @disable_trigger = 'disable' THEN
LEAVE customer_trigger;
END IF;IF (@v_cp_i' at line 2 [for Statement "LOCK TABLES `db`.`c_s_t` WRITE;
DROP TRIGGER IF EXISTS `db`.`c_s_t_aupd`;
CREATE TRIGGER `pt_osc_db_c_s_t_upd` AFTER UPDATE ON `db`.`c_s_t` FOR EACH ROW
customer_trigger: beginIF @disable_trigger = 'disable' THEN
LEAVE customer_trigger;
END IF;IF (@v_cp_id <> new.c_p_id) || (@v_cp_id IS NULL) THEN
set @v_cp_id=new.c_p_id;
insert into `db`.`log`(table_name,action,pk_id) values ('c_s_t','update',concat_ws(' ','c_p_id',new.c_p_id));
END IF;
REPLACE INTO `db`.`_c_s_t_new` (`c_p_id`, `p_s_id`, `status`, `created`, `modified`, `deleted`) VALUES (NEW.`c_p_id`, NEW.`p_s_id`, NEW.`status`, NEW.`created`, NEW.`modified`, NEW.`deleted`);
END;
UNLOCK TABLES;"] at ./bin/pt-online-schema-change line 10782.

I will try to fix this bug myself.

michaeldg:
Your trim() works multiline, so it will remove heading and trailing spaces on each line which is incorrect. This caused this issue. 

This is the change:
commit 333cb84db450e0eb90ad16c7b9398bf30f65f998
Author: Michaël de Groot <michael@intoparty.com>
Date:   Wed Jun 8 17:51:52 2016 -0500

    Fixed issue with trim() removing too much whitespace

diff --git a/bin/pt-online-schema-change b/bin/pt-online-schema-change
index 9438b75..584e26a 100755
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -10606,6 +10606,7 @@ sub create_triggers {
       # find if trigger has begin/end - supply insertion point
       my $delete_trigger_ins_point = trigger_ins_point(trigger => $orig_delete_trigger->[3]);
 
+
       $delete_trigger
          = "CREATE TRIGGER `${prefix}_del` AFTER DELETE ON $orig_tbl->{name} "
          . "FOR EACH ROW";
@@ -10637,6 +10638,7 @@ sub create_triggers {
          . "UNLOCK TABLES;";
       }
 
+
       my $definer = $orig_delete_trigger->[2];
       $definer =~ s/@/`@`/;
       $definer = "`".$definer."`";
@@ -10725,7 +10727,6 @@ sub create_triggers {
          = "LOCK TABLES $orig_tbl->{name} WRITE;\n"
          . "DROP TRIGGER IF EXISTS `$orig_tbl->{db}`.`$orig_update_trigger->[1]`;\n"
          . $update_trigger . "\n"
-         . terminate_sql(substr($orig_update_trigger->[3], 0, $update_trigger_ins_point)) . "\n"
          . $update_trigger_action . ";\n"
          . "END;\n"
          . "UNLOCK TABLES;";
@@ -11108,7 +11109,7 @@ sub terminate_sql {
 sub trim {
    my ( $text ) = @_;
    die "I need a text argument" unless defined $text;
-   $text =~ s/^\s+|\s+$//gm;
+   $text =~ s/^\s+|\s+$//g;
    return $text;
 }

ManjotS:
It needs to be multiline as in my testing there were some triggers that 
had empty lines after an END

michaeldg:
Why not remove empty lines on the end then instead?

michaeldg:
The multiline is not needed at all, empty lines are whitespaces as well in pcre:

root@ui-oltp01:~/percona-toolkit# cat test.pl 
$a="hello\n\n\n\n";  
print $a;
print "---------------\n";
$a=~s/^\s+|\s+$//g; 
print $a;
print "\nend\n";
root@ui-oltp01:~/percona-toolkit# perl test.pl 
hello



---------------
hello
end

PR #93, opened 2016-06-09
BUG-1585412 Fixed Date parsing regex for MySQL 5.7

percona-csalguero:
MySQL 5.7 general log dates are in RFC3339 format: 2006-01-02T15:04:05Z07:00
The previous version of the log parser was not able to parse this format for general log.

### DIFF
--- a/bin/pt-query-digest
+++ b/bin/pt-query-digest
@@ -9847,7 +9847,7 @@ sub new {
 
 my $genlog_line_1= qr{
    \A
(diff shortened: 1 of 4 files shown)


gtl-rahulsingh:
I am still not getting results from pt-query-digest(2.2.18 version) from slow query logs of mysql 5.7.12.basically since the timestamp format changed in slow query logs.

I didnt understand the solution if its mentioned somewhere in this thread.
Can any body please guide

PR #95, opened 2016-06-09
Bug-1402776

percona-csalguero:
The problem was the regex that parses the event data, including the db name, was looking for db + null at the end of the string but we have db + null + mysql_native_password

Launchpad: https://bugs.launchpad.net/percona-toolkit/+bug/1402776


frank-cizmich:
This change should go in MySQLProtocolParser.pm, and then updated in the tool.
Check that there are no discrepancies in the original tool with the update other than that which you change.

PR #102, opened 2016-06-22
add enum column type to is_char check so that values are properly quoted

dankinon:
My particular use case is that I have an enum value "0x0" which gets caught up in the hexadecimal check in https://github.com/percona/percona-toolkit/blob/2.2/bin/pt-table-sync#L1895 .  When this occurs, pt-table-sync treats the value as a hexadecimal value and bombs out when I get a "Data truncated for column" mysql error on insert/update.

### DIFF
--- a/bin/pt-table-sync
+++ b/bin/pt-table-sync
@@ -3537,7 +3537,7 @@ sub make_UPDATE {
    my $types = $self->{tbl_struct}->{type_for};
    return "UPDATE $self->{dst_db_tbl} SET "
       . join(', ', map {
-            my $is_char  = ($types->{$_} || '') =~ m/char|text/i;
+            my $is_char  = ($types->{$_} || '') =~ m/char|text|enum/i;
             my $is_float = ($types->{$_} || '') =~ m/float|double/i;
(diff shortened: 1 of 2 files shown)


percona-csalguero:
Thanks for this fix. It looks good and all tests are passing `All tests successful.
Files=45, Tests=234, 209 wallclock secs ( 0.15 usr  0.03 sys + 15.41 cusr  2.66 csys = 18.25 CPU)
Result: PASS
` but before merging we need to implement this change in `lib/ChangeHandler.pm` (line 329) as well and we also need a test case.

dankinon:
ChangeHandler.pm change is in.  The test case will be any enum value that starts with "0x", is that enough information or is there something more I can do to implement a test case?  This is my first real open source contribution so not sure what is expected of me.

percona-csalguero:
Don't worry. I'll write the test for you

dankinon:
Great, thanks!  Let me know if you need anything else.

percona-csalguero:
@dankinon Here is the PR I've made with the tests for your code: https://github.com/percona/percona-toolkit/pull/115

PR #116, opened 2016-07-14
Update NibbleIterator in pt-online-schema-change

dveeden:
NibbleIterator was updated in 47cc4225d82261a532dcda7c41d1ea9947b3bb89
but the change wasn't put into pt-online-schema-change.

I just ran ./util/update-modules bin/pt-online-schema-change


dveeden:
cc @frank-cizmich

percona-csalguero:
Thanks for the fix.

PR #121, opened 2016-08-01
Handle GTID ranges where the left-side integer is larger than 9

sodabrew:


percona-csalguero:
ok
All tests successful.
Files=1, Tests=7, 83 wallclock secs ( 0.01 usr  0.00 sys +  0.69 cusr  0.28 csys =  0.98 CPU)
Result: PASS

PR #122, opened 2016-08-01
pt-slave-restart version 2.2.18 is version 2.2.18

sodabrew:


sodabrew:
It's a nitpick, but was confusing me while I was debugging #121 and #123 on a production box and wasn't sure which version was installed :(

PR #123, opened 2016-08-01
Remove extra word 'default' from the --verbose help, it was confusing the option parser

sodabrew:


sodabrew:
Current release version:

```
$ bin/pt-slave-restart --help | grep verbose
  --quiet            -q  Suppresses normal output (disables --verbose)
  --verbose          -v  Adds more information to the default output (default 1)
  --verbose              output (default 1
```

Note that the default value for `--verbose` is not an integer, which causes any code that compares `$o->get('verbose')` against another integer to fail.

With this fix:

```
$ bin/pt-slave-restart --help | grep verbose
  --quiet            -q  Suppresses normal output (disables --verbose)
  --verbose          -v  Adds more information to the output (default 1)
  --verbose              1
```

percona-csalguero:
Thanks for this fix @sodabrew

PR #163, opened 2017-02-11
PT-2133 - added basic support for MongoDB to pt-stalk

fipar:
This branch adds basic MongoDB support to pt-stalk, adding the --mongo option, which: 
- Disables the MySQL collectors, 
- Collects db.currentOp(), db.isMaster(), sh.status(), rs.status() and db.serverStatus()
- One of sh.status or rs.status will fail as they are mutually exclusive, but the collect_mongo function does not check this and runs both anyway
- Provides trg_mongo_default as the default mongo trigger (count of 'inprog' current operations) if --function is not specified

It includes no tests of these functions as it does not provide a mongodb sandbox, but it does not break any of the existing pt-stalk tests.

### DIFF
--- a/bin/pt-stalk
+++ b/bin/pt-stalk
@@ -786,8 +786,19 @@ CMD_SYSCTL="${CMD_SYSCTL:-"$(_which sysctl)"}"
 CMD_TCPDUMP="${CMD_TCPDUMP:-"$(_which tcpdump)"}"
 CMD_VMSTAT="${CMD_VMSTAT:-"$(_which vmstat)"}"
 CMD_DMESG="${CMD_DMESG:-"$(_which dmesg)"}"
+CMD_MONGO="${CMD_MONGO:-"$(_which mongo)"}"
 
 [ -z "$CMD_SYSCTL" -a -x "/sbin/sysctl" ] && CMD_SYSCTL="/sbin/sysctl"
+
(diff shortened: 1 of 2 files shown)


dbmurphy:
This should be moved to the new PT in go tools

svetasmirnova (changes requested):
I am sorry it took too long to review this but I have a few suggestions.

1. I believe that instead of having MySQL and MongoDB collections mutually exclusive, it is better to not mix up MySQL and MongoDB connection options, rather collect their data independently.
2. Since we have option --system-only now, maybe it makes sense to implement more complicated logic and let users choose if they want to collect: system, mysql and mongo info, only system, only system and mysql, only system and mongo, only mysql, only mongo, and (why not?) only mysql and mongo
3. Please add a test case. You will find examples in t/pt-stalk, t/lib/bash/collect.t, and t/lib/bash/collect.sh

dbmurphy:
That seems overly complex maybe system and db-engine should be independent.

`--system --db-engine mongo` where system is a flag and db-engine is a list with specific connection variables prefixed for each engine (keeping no prefix assuming MySQL and system-only mapped to `--system --db-engine none`

Really we should be talking about a parent command which calls one of N child commands based on db-engine. But that would be a much bigger redesign but more long term scalable

PR #169, opened 2017-02-23
PT-73 Added support for SSL

percona-csalguero:


rnkhouse:
how to provide ssl-ca from command line? there is no option. @percona-csalguero

PR #193, opened 2017-03-23
Turn off statement based binlog checks

jacobbednarz:
For RDS we use row based replication however `pt-table-checksum` wants
to enforce statement and despite passing in the
`--no-check-binlog-format` flag, it would still run it (even though the
documentation [suggests](https://www.percona.com/doc/percona-toolkit/3.0/pt-table-checksum.html#cmdoption-pt-table-checksum--[no]check-binlog-format) it shouldn't). This appears to be a bug
because there is another binlog format check that is wrapped in
`$o->get('check-binlog-format')` slightly further down.

To resolve this, I've updated the remaining binlog format check in the
same expression as the one below and it now honours the
`--no-check-binlog-format` as expected.

### DIFF
--- a/bin/pt-table-checksum
+++ b/bin/pt-table-checksum
@@ -9272,35 +9272,36 @@ sub main {
          die "Error setting SQL_MODE"
            . ": $EVAL_ERROR";
       }
- 
 
-      # https://bugs.launchpad.net/percona-toolkit/+bug/919352
-      # The tool shouldn't blindly attempt to change binlog_format;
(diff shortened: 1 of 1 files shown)


jacobbednarz:
Recommend using `?w=1` on the diff to avoid a bit of confusion.

percona-csalguero:
ok
All tests successful.
Files=328, Tests=4430, 1249 wallclock secs ( 1.45 usr  0.31 sys + 105.24 cusr 20.82 csys = 127.82 CPU)
Result: PASS

PR #209, opened 2017-04-26
PMM-720: encapsulate stats, separate them from profiler

arvenil:
Profiler program is separated from actually calculating stats. Also encapsulating stats allows to avoid ugly `ProcessDoc(doc, stats)` that was affecting map passed by reference.

So, now we have
```go
type Stats interface {
	Add(doc proto.SystemProfile) error // adds doc to collection
	Queries() stats.Queries            // returns processed collection
	Reset()                            // resets/deletes collection
}
```

full implementation

```go
// New creates new instance of stats with given fingerprinter
func New(fingerprinter fingerprinter.Fingerprinter) *Stats {
	s := &Stats{
		fingerprinter: fingerprinter,
	}

	s.Reset()
	return s
}

// Stats is a collection of MongoDB statistics
type Stats struct {
	// dependencies
	fingerprinter fingerprinter.Fingerprinter

	// internal
	queryInfoAndCounters map[GroupKey]*QueryInfoAndCounters
}

// Reset clears the collection of statistics
func (s *Stats) Reset() {
	s.queryInfoAndCounters = make(map[GroupKey]*QueryInfoAndCounters)
}

// Add adds proto.SystemProfile to the collection of statistics
func (s *Stats) Add(doc proto.SystemProfile) error {
	fp, err := s.fingerprinter.Fingerprint(doc.Query)
	if err != nil {
		return fmt.Errorf("cannot get fingerprint: %s", err)
	}
	var qiac *QueryInfoAndCounters
	var ok bool

	key := GroupKey{
		Operation:   doc.Op,
		Fingerprint: fp,
		Namespace:   doc.Ns,
	}
	if qiac, ok = s.queryInfoAndCounters[key]; !ok {
		realQuery, _ := util.GetQueryField(doc.Query)
		qiac = &QueryInfoAndCounters{
			ID:          fmt.Sprintf("%x", md5.Sum([]byte(fmt.Sprintf("%s", key)))),
			Operation:   doc.Op,
			Fingerprint: fp,
			Namespace:   doc.Ns,
			TableScan:   false,
			Query:       realQuery,
		}
		s.queryInfoAndCounters[key] = qiac
	}
	qiac.Count++
	qiac.NScanned = append(qiac.NScanned, float64(doc.DocsExamined))
	qiac.NReturned = append(qiac.NReturned, float64(doc.Nreturned))
	qiac.QueryTime = append(qiac.QueryTime, float64(doc.Millis))
	qiac.ResponseLength = append(qiac.ResponseLength, float64(doc.ResponseLength))
	var zeroTime time.Time
	if qiac.FirstSeen == zeroTime || qiac.FirstSeen.After(doc.Ts) {
		qiac.FirstSeen = doc.Ts
	}
	if qiac.LastSeen == zeroTime || qiac.LastSeen.Before(doc.Ts) {
		qiac.LastSeen = doc.Ts
	}

	return nil
}

// Queries returns all collected statistics
func (s *Stats) Queries() Queries {
	return mapToArray(s.queryInfoAndCounters)
}

// Queries is a slice of MongoDB statistics
type Queries []QueryInfoAndCounters

// CalcQueriesStats calculates QueryStats for given uptime
func (q Queries) CalcQueriesStats(uptime int64) []QueryStats {
	qs := []QueryStats{}
	tc := calcTotalCounters(q)

	for _, query := range q {
		queryStats := countersToStats(query, uptime, tc)
		qs = append(qs, queryStats)
	}

	return qs
}

// CalcTotalQueriesStats calculates total QueryStats for given uptime
func (q Queries) CalcTotalQueriesStats(uptime int64) QueryStats {
	tc := calcTotalCounters(q)

	totalQueryInfoAndCounters := aggregateCounters(q)
	totalStats := countersToStats(totalQueryInfoAndCounters, uptime, tc)

	return totalStats
}
```

### DIFF
--- a/src/go/mongolib/profiler/profiler.go
+++ b/src/go/mongolib/profiler/profiler.go
@@ -1,141 +1,63 @@
 package profiler
 
 import (
-	"crypto/md5"
-	"encoding/json"
-	"errors"
-	"fmt"
 	"sync"
 	"time"
 
-	"github.com/montanaflynn/stats"
-	"github.com/percona/percona-toolkit/src/go/mongolib/fingerprinter"
 	"github.com/percona/percona-toolkit/src/go/mongolib/proto"
-	"github.com/percona/percona-toolkit/src/go/mongolib/util"
+	"github.com/percona/percona-toolkit/src/go/mongolib/stats"
 	"github.com/percona/percona-toolkit/src/go/pt-mongodb-query-digest/filter"
 	"github.com/percona/pmgo"
 )
 
 var (
-	// MaxDepthLevel Max recursion level for the fingerprinter
-	MaxDepthLevel = 10
 	// DocsBufferSize is the buffer size to store documents from the MongoDB profiler
 	DocsBufferSize = 100
-	// ErrCannotGetQuery is the error returned if we cannot find a query into the profiler document
-	ErrCannotGetQuery = errors.New("cannot get query field from the profile document (it is not a map)")
 )
-
-// Times is an array of time.Time that implements the Sorter interface
-type Times []time.Time
-
-func (a Times) Len() int           { return len(a) }
-func (a Times) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
-func (a Times) Less(i, j int) bool { return a[i].Before(a[j]) }
-
-type StatsGroupKey struct {
-	Operation   string
-	Fingerprint string
-	Namespace   string
-}
-
-type totalCounters struct {
-	Count     int
-	Scanned   float64
-	Returned  float64
-	QueryTime float64
-	Bytes     float64
-}
 
 type Profiler interface {
 	GetLastError() error
-	QueriesChan() chan []QueryInfoAndCounters
+	QueriesChan() chan stats.Queries
 	TimeoutsChan() <-chan time.Time
-	ProcessDoc(proto.SystemProfile, map[StatsGroupKey]*QueryInfoAndCounters) error
 	Start()
 	Stop()
 }
 
 type Profile struct {
-	filters      []filter.Filter
-	iterator     pmgo.IterManager
-	ticker       <-chan time.Time
-	queriesChan  chan []QueryInfoAndCounters
+	// dependencies
+	iterator pmgo.IterManager
+	filters  []filter.Filter
+	ticker   <-chan time.Time
+	stats    Stats
+
+	// internal
+	queriesChan  chan stats.Queries
 	stopChan     chan bool
 	docsChan     chan proto.SystemProfile
 	timeoutsChan chan time.Time
 	// For the moment ProcessDoc is exportable to it could be called from the "outside"
 	// For that reason, we need a mutex to make it thread safe. In the future this func
 	// will be unexported
-	countersMapLock        sync.Mutex
-	queriesInfoAndCounters map[StatsGroupKey]*QueryInfoAndCounters
-	keyFilters             []string
-	fingerprinter          fingerprinter.Fingerprinter
-	lock                   sync.Mutex
-	running                bool
-	lastError              error
-	stopWaitGroup          sync.WaitGroup
+	countersMapLock sync.Mutex
+	keyFilters      []string
+	lock            sync.Mutex
+	running         bool
+	lastError       error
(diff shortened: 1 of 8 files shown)


src/go/mongolib/util/main_test.go
```diff
@@ -145,9 +145,7 @@ func TestGetReplicasetMembers(t *testing.T) {
 	database.EXPECT().Run(bson.D{{"serverStatus", 1}, {"recordStats", 1}}, gomock.Any()).SetArg(1, ss)
 	session.EXPECT().Close()
 
-	session.EXPECT().Close()
```
arvenil: This test were broken before my changes, I fixed them. It looks like `runtests.sh` script is broken, I didn't fixed it. https://github.com/percona/percona-toolkit/blob/3.0/src/go/runtests.sh#L7-L12 Script lists only top level packages so e.g. `mongolib` but not subpackages e.g. `mongolib/util` and then it runs tests only for that package with `go test` (alternative would be `go test ./...` but this doesn't work for coverage).
travis.ci would be nice to have too :P

src/go/mongolib/profiler/profiler.go
```diff
@@ -217,163 +140,9 @@ func (p *Profile) getDocs() {
...
-
-	return nil
-
-}
-
-func CalcQueriesStats(queries []QueryInfoAndCounters, uptime int64) []QueryStats {
-	stats := []QueryStats{}
-	tc := calcTotalCounters(queries)
-
-	for _, query := range queries {
```
arvenil: Do we really need to convert map to slice before passing them to `Calc*` funcs? You can `range` over a map so I'm not sure why to do this. Without this conversion we could have `Calc*` funcs directly on `Stats` and avoid unnecessary rewrite loop. It's a quick thought so I might be totally wrong :P

percona-csalguero: Because we need to show them in order for pt-mongodb-query-digest. They are ranked from most used to less used, etc. I think we can de-couple this.

src/go/mongolib/stats/stats.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+	var qiac *QueryInfoAndCounters
+	var ok bool
+
+	key := GroupKey{
+		Operation:   doc.Op,
+		Fingerprint: fp,
+		Namespace:   doc.Ns,
+	}
+	if qiac, ok = s.queryInfoAndCounters[key]; !ok {
```
percona-csalguero: Just a question without analyzing all the flow: don't we need a mutex here? Add is an exported function and it could be called concurrently but Go maps are not thread safe.

src/go/mongolib/stats/stats.go
```diff
@@ -0,0 +1,268 @@
...
+// Queries returns all collected statistics
+func (s *Stats) Queries() Queries {
+	return mapToArray(s.queryInfoAndCounters)
+}
+
+// Queries is a slice of MongoDB statistics
+type Queries []QueryInfoAndCounters
+
+// CalcQueriesStats calculates QueryStats for given uptime
+func (q Queries) CalcQueriesStats(uptime int64) []QueryStats {
```
percona-csalguero: We can remove or rename this since this function is only used by pt-mongodb-query-digest

src/go/mongolib/stats/stats.go
arvenil: True, I will add it, I remember I removed it and forgot to re add :P Good catch :)

src/go/mongolib/profiler/profiler.go
arvenil: After rethinking this code is ok actually 👍

src/go/mongolib/stats/stats.go
arvenil: It's ok, I'm using this one, I'm not using the other one but still I think this is quite correct place for this functions. 👍

src/go/mongolib/stats/stats.go
arvenil: Done

percona-csalguero:
:+1:

arvenil:
Should I merge it then?

PR #213, opened 2017-05-08
PT-135 - Only return slaves which have a port

bbeaudreault:
Some binlog dump connections (not normal replicas) may not be using a port. pt-online-schema-change picks those up because of the use of `SHOW FULL PROCESSLIST`, but in those cases, the tool fails with the following:

```
Use of uninitialized value $host in string eq at /usr/bin/pt-online-schema-change line 4256, <STDIN> line 1.
```

Replication never uses sockets, but non-replication users of Binlog Dump may. So for our purposes we should require a port. This change filters the list of returned slaves to only those which have the `:` we are filtering on in `_find_slaves_by_processlist`.

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -4332,7 +4332,7 @@ sub get_connected_slaves {
 
    $sql = 'SHOW FULL PROCESSLIST';
    PTDEBUG && _d($dbh, $sql);
-   grep { $_->{command} =~ m/Binlog Dump/i }
+   grep { $_->{command} =~ m/Binlog Dump/i && $_->{host} =~ m/^([^:]+):/ }
    map  { # Lowercase the column names
       my %hash;
       @hash{ map { lc $_ } keys %$_ } = values %$_;


bbeaudreault:
I'd actually like to get this fixed in 2.2 -- I accidentally submitted this against 3.0 though. Ideally it should be fixed for both, so if this looks good I can resubmit against 2.2 afterwards.

bbeaudreault:
I've created a JIRA for this https://jira.percona.com/browse/PT-135

percona-csalguero:
Could you add a test case please?

it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-toolkit?pullRequest=213) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.percona.com:443/percona/percona-toolkit?pullRequest=213) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.percona.com:443/check/percona/percona-toolkit?pullRequest=213) it.</sub>

PR #255, opened 2017-08-26
pt-query-digest ：--create-history-table out of action

kaerdo:
create-history-table，create-review-table   ||  Yes or no out of action  
Cause a lot of “ CREATE DATABASE/TABLE IF NOT EXISTS” ,lot of metadata_locks
pt-query-digest version  3.0.4
mysq version :5.7


svetasmirnova:

According to the comment:

> \# Even if the db already exists, do this in case it does not exist
> \# on a slave.

These CREATE ... IF NOT EXIST statements are intentional. Also, these is just two tables, and you can avoid creating them by passing options --no-create-history-table and --no-create-review-table

PR #259, opened 2017-08-31
Add .travis.yml

arvenil:
Adds .travis.yml for go part of the project. It introduces matrix of MongoDB instances - you are not using such tests anywhere, but I do in one of my PRs so I've already introduced that. Let me know what you think.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,54 @@
+dist: trusty
+sudo: required
+
+language: go
+
+go:
+  - 1.8.x
+
+services:
+  - docker
+
+env:
+  matrix:
+    - MONGODB_IMAGE=mongo:2.6
+    - MONGODB_IMAGE=mongo:3.0
+    - MONGODB_IMAGE=mongo:3.2
+    - MONGODB_IMAGE=mongo:3.4
+    - MONGODB_IMAGE=percona/percona-server-mongodb:3.0
+    - MONGODB_IMAGE=percona/percona-server-mongodb:3.2
+    - MONGODB_IMAGE=percona/percona-server-mongodb:3.4
+
+matrix:
+  include:
+    - go: tip
+      env:
+
+install:
+  - go get -u github.com/Masterminds/glide
+  # remove vendor dir and re-fetch it to check later if it's correct with `git diff --exit-code`
+  - rm -rf vendor/
+  - glide install
+  # check if vendor dir is correct
+  - git diff --exit-code
+
+before_script:
+  # run docker containers
+  - docker-compose up -d
+  # log versions
+  - docker --version
+  - docker-compose --version
+  - docker-compose exec mongo mongo --version
+  # we need at least one document in test db
+  - mongo --eval 'db.init.insert({})'
+
+script:
+  - go test -timeout 1m $(glide nv)
+
+notifications:
+  email: false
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -0,0 +1,54 @@
...
+    - MONGODB_IMAGE=mongo:3.0
+    - MONGODB_IMAGE=mongo:3.2
+    - MONGODB_IMAGE=mongo:3.4
+    - MONGODB_IMAGE=percona/percona-server-mongodb:3.0
+    - MONGODB_IMAGE=percona/percona-server-mongodb:3.2
+    - MONGODB_IMAGE=percona/percona-server-mongodb:3.4
+
+matrix:
+  include:
+    - go: tip
```
AlekSi: Can we also add `1.9.x`?

.travis.yml
```diff
@@ -0,0 +1,54 @@
...
+      env:
+
+install:
+  - go get -u github.com/Masterminds/glide
+  # remove vendor dir and re-fetch it to check later if it's correct with `git diff --exit-code`
+  - rm -rf vendor/
+  - glide install
```
AlekSi: `glide install -v` to strip nested vendor dirs?

.travis.yml
```diff
@@ -0,0 +1,54 @@
...
+  - docker-compose up -d
+  # log versions
+  - docker --version
+  - docker-compose --version
+  - docker-compose exec mongo mongo --version
+  # we need at least one document in test db
+  - mongo --eval 'db.init.insert({})'
+
+script:
+  - go test -timeout 1m $(glide nv)
```
AlekSi: Soon, very soon we will switch to Go 1.9 and get rid of that `nv` nonsense 😂

src/go/mongolib/proto/main_test.go
```diff
@@ -36,12 +34,11 @@ func TestMain(m *testing.M) {
...
-	if err := session.DB("admin").Run(bson.D{{"serverStatus", 1}, {"recordStats", 1}}, &ss); err != nil {
+func ExamplePing() {
+	ss := map[string]interface{}{}
+	if err := session.DB("admin").Run(bson.D{{"ping", 1}}, &ss); err != nil {
 		panic(err)
 	}
 	fmt.Printf("%+v", ss)
-	// Output:
-
+	// Output: map[ok:1]
```
arvenil: @cfsalguero I've changed that test because `serverStatus` gives result depending on machine it was run on. In other words it was giving different result on my local host and travis etc. Let me know if you think it should look different.

src/go/mongolib/stats/stats.go
```diff
@@ -110,9 +111,15 @@ func (s *Stats) Queries() Queries {
...
+	}
+	sort.Sort(keys)
+
 	queries := []QueryInfoAndCounters{}
-	for _, v := range s.queryInfoAndCounters {
-		queries = append(queries, *v)
+	for _, key := range keys {
+		queries = append(queries, *s.queryInfoAndCounters[key])
 	}
 	return queries
```
arvenil: @cfsalguero queries were converted from unsorted map to sorted slice, so the order of elements was unpredictable, by sorting them by key I've kept the order. Let me know if you want this different.

PR #263, opened 2017-09-13
pt-mongodb-query-digest improvements

matsuu:
- support versions less than 3.2.0
- fix incorrect output


percona-csalguero:
Thanks for the fixes.

PR #271, opened 2017-10-10
Fix have innodb on 57

fipar:
have_innodb is used to decided if the Innodb section of pt-mysql-summary's report is printed, and it is no longer available on 5.7. This uses innodb_buffer_pool_size instead.

### DIFF
--- a/bin/pt-mysql-summary
+++ b/bin/pt-mysql-summary
@@ -1374,7 +1374,7 @@ pretty_print_cnf_file () {
 
    perl -n -l -e '
       my $line = $_;
-      if ( $line =~ /^[ \t]*[a-zA-Z[]/ ) { 
+      if ( $line =~ /^\s*[a-zA-Z[]/ ) { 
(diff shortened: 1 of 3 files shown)


bin/pt-mysql-summary
```diff
@@ -2162,41 +2150,6 @@ parse_wsrep_provider_options () {
    ' "$looking_for"
 }
 
-report_jemalloc_enabled() {
```
percona-csalguero: why this has been removed?

fipar: Seems like an error in a merge as I did not change that. I'll see if I can find when that happened, or will submit a new PR with only the have_innodb changes on it.

fipar: The problem is introduced when I run update-modules for pt-mysql-summary, so I'll close this PR, create a new one changing just the module, and let you check if my change is causing that to break (I could not find other references to have_innodb but I could be wrong) or if you can update pt-mysql-summary yourself based on the updated module. Thanks!

PR #287, opened 2017-12-18
fix for broken slave-arguments in pt-table-sync & pt-table-checksum

Sadrak:
Hello,

we have a special replication setup and i run into this bugs.

Have a nice day
Felix

### DIFF
--- a/bin/pt-table-checksum
+++ b/bin/pt-table-checksum
@@ -5223,7 +5223,7 @@ sub recurse_to_slaves {
          PTDEBUG && _d('Recursing from',
             $dp->as_string($dsn), 'to', $dp->as_string($slave));
          $self->recurse_to_slaves(
-            { %$args, dsn => $slave, dbh => undef, parent => $dsn, slave_user => $slave_user, $slave_password => $slave_password }, $level + 1 );
+            { %$args, dsn => $slave, dbh => undef, parent => $dsn, slave_user => $slave_user, slave_password => $slave_password }, $level + 1 );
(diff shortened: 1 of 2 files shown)


Sadrak:
uh, the second commit should be another pull request ... sorry for that failure.

Explanation:
We have a backup-server with 4 slaves (4 mysql instances) on different ports with different masters.
The available connection option don't work for us. When we connect to the master, the slave connection won't work because of the wrong port. When we connect to the slave, the master connection won't work because of different user & password.

I think it would be better to allow more manually options so someone can define the slave & master connection in both tools (and the tool only verify if it is sane).

Like in pt-table-checksum the multi DSN option (but there are some problems too).

Sadrak:
Another bug:
When you use the defaults-file (F=) dsn option, the file can contain a hostname. But the tools always set a host=localhost; in the dsn if no host was specified and that override the hostname from the defaults-file.

percona-csalguero:
I think the correct fix would be to add support for replication method=dsns since adding `slave_host` parameter would allow to specify only one slave.

I know it is not easy because pt-table-checksum uses the `recurse_to_slaves` function and it is not easy to implement dsns for it.

svetasmirnova:
Recursion method DSNS is available for pt-table-checksum and pt-table-sync: https://docs.percona.com/percona-toolkit/pt-table-checksum.html#cmdoption-pt-table-checksum-recursion-method, and this is, definitely, better method. So closing this PR.

PR #297, opened 2018-01-25
PT-1509 Only set binlog_format when necessary

moritz:
The documentation states that --sync-to-master or --replicate need
the `binlog_format` to be `STATEMENT`, but the code set the `binlog format`
regardless of the operation mode.

With this patch, the `binlog_format` is only set if `--sync-to-master`
or `--replicate` options have actually been supplied on the command line,
enabling a node-to-node sync without these options without the SUPER
privilege.

### DIFF
--- a/bin/pt-table-sync
+++ b/bin/pt-table-sync
@@ -10957,7 +10957,8 @@ sub get_cxn {
    # instead, it should check if it's already set to STATEMENT.
    # This is becase starting with MySQL 5.1.29, changing the format
    # requires a SUPER user.
-   if ( VersionParser->new($dbh) >= '5.1.29' ) {
+   if ( VersionParser->new($dbh) >= '5.1.29'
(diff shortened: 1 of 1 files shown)


moritz:
FWIW I've signed a CLA about an hour ago, but the bot doesn't seem to pick up on it. Can somebody please check that manually? (also applies to #298)

moritz:
Any comment from @percona-csalguero or any other contributor would be appreciated. Thanks!

percona-csalguero:
Created https://jira.percona.com/browse/PT-1509 to track this PR

percona-csalguero:
```
All tests successful.
Files=47, Tests=239, 192 wallclock secs ( 0.12 usr  0.04 sys + 21.32 cusr  5.50 csys = 26.98 CPU)
Result: PASS
```

peterschwaller:
Dummy comment to trigger sync to Jira.  Please ignore.

PR #301, opened 2018-02-12
PMM-1905: fix "NToReturn value must be non-negative" issue.

arvenil:
* PMM-1905
* cleanup data samples
* switch from testing mongo 3.5 to 3.6


arvenil:
@cfsalguero done

PR #319, opened 2018-03-27
Add --[no]-collect-mysql

fipar:
This PR adds the --collect-mysql option, which is enabled by default. 
If disabled (by specifying --no-collect-mysql) then none of the MySQL collectors (or checks, such as presence of mysql and mysqladmin on the path) are performed. 

The goal is to make it possible to use pt-stalk with plugins that capture data about other systems (e.g. Cassandra) without having it failed due to missing mysql binaries on the host.

### DIFF
--- a/bin/pt-stalk
+++ b/bin/pt-stalk
@@ -797,7 +797,7 @@ collect() {
    local p="$2"  # prefix for each result file
 
    local mysqld_pid=""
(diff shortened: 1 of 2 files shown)


svetasmirnova:
Implemented slightly different as part of fix for https://jira.percona.com/browse/PT-2037

PR #359, opened 2018-08-14
Fix PT-1605, 3.0 version, documenting data collected... 

fiowro:
with the --version-check option


dbpercona:
I reviewed the documentation as published to the test site and it looks good.

PR #369, opened 2018-09-25
Patch newlines in table columns

geneguido:
mysql 5.6.40 allows newlines in column names however the following code:

``` my @defs = $ddl =~ m/^(\s+`.*?),?$/gm; ```

breaks due to it detecting newlines as line ends. The 'm' argument at the end does this by auto-detecting lines by newline characters.

To correct this issue I've made use of zero-length assertions known as " positive lookback"

https://www.regular-expressions.info/lookaround.html

what does it do?

```m/(?:(?<=,\n)|(?<=\(\n))(\s+`(?:.|\n)+?`.+?),?\n/g; ```

TLDR:

Treat the string as one long string and don't treat \n as the end of a line.

look for ``` (\s+`(?:.|\n)+?`.+?),?\n ```

if one of those matches look at what precedes the string

if it's ',\n' or '(\n' the string matches. Only save what's in ``` (\s+`(?:.|\n)+?`.+?) ```

m/ is declaring this a matching regex.

``` (?:(?<=,\n)|(?<=(\n)) ``` This is an OR statement including two look-behind clauses. The ?: tells the enclosing parentheses to not store the result as a variable. I've put the two look-behinds in this OR statement below this line:

(?<=,\n) Look behind the matched string for a comma followed by a newline, the comma must be there for this look behind to match.

(?<=(\n) Look behind the matched string for a open parentheses followed by a newline, the open parentheses must be there.

``` (\s+`(?:.|\n)+?`.+?),?\n ``` This is the actual match. Match newline character followed by one or more spaces followed by back-tick followed by a character which can be any character or a newline one or more times, but don't be greedy and take the rest of the match into consideration. Followed by a back tick and any character one or more times. This match stops where there is a comma or failing that a newline following a back tick and some characters.

,?\n match a comma that may not be there followed by a newline.
/g don't stop if this pattern matches keep looking for more patterns to the end of the string.

### DIFF
--- a/bin/pt-table-checksum
+++ b/bin/pt-table-checksum
@@ -4481,7 +4481,7 @@ sub parse {
 
    my $engine = $self->get_engine($ddl);
 
-   my @defs   = $ddl =~ m/^(\s+`.*?),?$/gm;
+   my @defs   = $ddl =~ m/(?:(?<=,\n)|(?<=\(\n))(\s+`(?:.|\n)+?`.+?),?\n/g;
    my @cols   = map { $_ =~ m/`([^`]+)`/ } @defs;
    PTDEBUG && _d('Table cols:', join(', ', map { "`$_`" } @cols));
 

### CODE REVIEW

**Review by percona-csalguero — CHANGES_REQUESTED** [review]
This change goes into the `TableParser` package and it should be applied to all programs using that pkg.
Can we also add a test?
Currently all tests for table checksum are passing but I would prefer to have a specific test.
```
All tests successful.
Files=37, Tests=256, 277 wallclock secs ( 0.17 usr  0.04 sys + 26.85 cusr  8.11 csys = 35.17 CPU)
Result: PASS
```

geneguido:
Anyone have a quick one-liner they use to run the tests? I got 
/usr/bin/perl -MExtUtils::Command::MM -MTest::Harness -e "undef *Test::Harness::Switches; test_harness(0, 'blib/lib', 'blib/arch')" t/*/*.t 

But every test fails so i don't think this is right.
After reading the make file and editing the way it globs test files.

I may have a test that works having taken the Bug 821675: can't parse column names containing periods, and copying to instead test for newlines. But I'll be honest, I am out of practice with this stuff and have no idea how to run the test suite to figure out if my test it good or not.

percona-csalguero:
In https://github.com/percona/percona-toolkit/blob/3.0/CONTRIBUTE.md#setting-up-the-development-environment you can find instructions about how to set up the sandbox and run the tests.

Thanks.

svetasmirnova (changes requested):
Code you changed is part of the TableParser package and originally defined in lib/TableParser.pm You need to put your changes there, otherwise they will be rewritten after we run util/update-modules

nogasgofast:
Sorry guys I ran out of time to work on this issue. And won't be coming back to it.

PR #377, opened 2018-12-04
PT-1662 fix pt-secure-collect doc styles

fiowro:


michaelcoburn:
- [ ] Add example for pt-secure-collect help
- [ ] Confirm these work –help-long and –help-man, and if so format them appropriately
- [ ] Under temp dir, format the following "${HOME}/data_collection_{timestamp}"
- [ ] Remove these two sentences "Internal variables placeholders will be replaced with the corresponding flag values. For example, $mysql-host will be replaced with the values specified in the --mysql-host flag."

fiowro:
I've applied all fixes beside the help command example: it looks close to the text on this page, and it will be good to avoid too much duplicated information. Btw, --help-long and --help-man work as expected.

PR #381, opened 2018-12-13
Fix slave-user and slave-password with pt-table-checksum

tobiaspal:
When using `pt-table-checksum` without `--sync-to-master` and the DSN as the master to sync all slaves, the `--slave-user` and `--slave-password` parameters are ignored.

Steps to reproduce:
1. Set up master-slave replication with percona-server-5.7
1. Create a user to be used as the `slave-user` with a random password:
    ```
    GRANT ALL PRIVILEGES ON *.* TO 'pt-table-checks' IDENTIFIED BY 'auW2phohsiey';
    ```
    This user will get replicated to the slaves.
1. Run `pt-table-checksum` to create the `percona.checksums` table:
    ```
    pt-table-checksum --slave-user=pt-table-checks --slave-password=auW2phohsiey
    ```
    This uses the `slave-user` parameter correctly and finishes successfully.
2. Run `pt-table-sync` on the master with the `slave-user` parameter to sync the slaves to the master:
    ```
    pt-table-sync --replicate=percona.checksums --dry-run localhost
    ```
    This fails with:
    ```
    Cannot connect to h=10.4.28.78
    ```
    Because it's not using the `slave-user` parameter to connect to the slave.


ChristianCiach:
Please merge this! I just wasted hours trying to find out why setting the "--slave-user" option has no effect.

svetasmirnova (changes requested):

You modified the auto-generated code that will be removed once we run the update-modules utility. Please put your changes into the MasterSlave package instead.

svetasmirnova:
Currently, you can provide a replica username and password as part of the DSN.

svetasmirnova:
Other issues with the fix are:

- options take precedence even if user and password defined in the dsns table. This could cause issues if users have, say, 10 replicas with the same user/password pairs and two with different ones.
- you access tool options in the library code
- there is no test case.

There is also a similar PR at https://github.com/percona/percona-toolkit/pull/425

svetasmirnova:
We will consider similar https://github.com/percona/percona-toolkit/pull/425 instead.

PR #230, opened 2017-06-09
Safely check for undefined values in --skip-check-slave-lag

cswingler:
The P parameter can be undefined in DSNs, resulting in the skip-check-slave-lag feature in pt-table-checksum puking trying to compare `undef`s.

This handles that check a little more safely.

PR #388, opened 2019-03-04
Check undefined skip check slave lag schema change

cswingler:
The P parameter can be undefined in DSNs, resulting in the skip-check-slave-lag feature in pt-table-checksum puking trying to compare undefs.

This handles that check a little more safely.

This is exactly like #230 , but for `pt-online-schema-change`.


hhgyu:
the issue still exists.

svetasmirnova (approved):

Please sign CLA, so we can accept it.

svetasmirnova:
We may need to check if other tools require similar fix.

svetasmirnova (approved):

EvgeniyPatlan (approved):

PR #399, opened 2019-05-16
Increase time precision

jouir:
Time type in OptionParser accepts only integers. In pt-kill, we can set `interval` option (time) to 0, to run as fast as possible, or 1 second, with the risk of being too slow. Say we have more than one connection that matches the pattern every second, pt-kill cannot keep up with the killing rate or will allocate a full CPU core all the time. If we insist and we set a float like 0.25, it's parsed as 25. Theses commits allow floating-point numbers to be passed for time. In pt-kill, the `sleep` function from `Time::HiRes` module has to be used because Perl built-in `sleep` function doesn't handle floating numbers.

### DIFF
--- a/bin/pt-kill
+++ b/bin/pt-kill
@@ -724,7 +724,7 @@ sub _validate_type {
 
    if ( $val && $opt->{type} eq 'm' ) {  # type time
       PTDEBUG && _d('Parsing option', $opt->{long}, 'as a time value');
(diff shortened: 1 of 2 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-toolkit?pullRequest=399) <br/>All committers have signed the CLA.

jouir:
A workaround is to use `--victims all` argument to kill all matching sessions at once instead of killing one after another and re-enter the loop (with interval). But it would be nice to have sub-second intervals.

PR #408, opened 2019-07-18
PBM-144 Updated MasterSlave to support hosts without port

percona-csalguero:


tplavcic:
@percona-csalguero PT-144 instead of PBM-144 ?

rnovikovP:
@tplavcic , @percona-csalguero  looks like  this is for PT-79

PR #424, opened 2019-10-16
Added option to log queries to database

sjalilund:
Added option to log potential victims to the database specified in the dsn that is specified for logging killed queries.
This to get a central place to store the logged potential victims, instead of just locally on the file system of the target database.

### DIFF
--- a/bin/pt-kill
+++ b/bin/pt-kill
@@ -1949,7 +1949,7 @@ sub parse {
    foreach my $key ( keys %$opts ) {
       PTDEBUG && _d('Finding value for', $key);
       $final_props{$key} = $given_props{$key};
-      if (   !defined $final_props{$key}
+      if ( !defined $final_props{$key}  
            && defined $prev->{$key} && $opts->{$key}->{copy} )
       {
          $final_props{$key} = $prev->{$key};
@@ -1974,6 +1974,10 @@ sub parse {
                . "for complete documentation."
             unless $final_props{$key};
       }
+   }
+
+   if ($final_props{F}) {
+       %final_props = ( 'F' => $final_props{F} );
    }
 
    return \%final_props;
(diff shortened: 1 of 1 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-toolkit?pullRequest=424) <br/>All committers have signed the CLA.

svetasmirnova (changes requested):

Please add a test case. You will find examples in the directory t.

bin/pt-kill
```diff
@@ -7379,6 +7379,17 @@ sub main {
...
                msg('Executed ' . $o->get('execute-command'));
             }
+            if ( $o->get('logdb') ) {
+               if ( $log ) {
+                  log_to_table(
+                     log        => $log,
+                     query      => $query,
+                     proclist   => $pl,
+                     columns    => \@processlist_columns,
+                     eval_error => $EVAL_ERROR,
```
svetasmirnova, line 7389: EVAL_ERROR makes sense only after pt-kill tried to kill the query.

bin/pt-kill
```diff
@@ -7379,6 +7379,17 @@ sub main {
                exec_cmd($o->get('execute-command'));
                msg('Executed ' . $o->get('execute-command'));
             }
+            if ( $o->get('logdb') ) {
+               if ( $log ) {
```
svetasmirnova, line 7383: What if someone specifies option --logdb and not --log?

bin/pt-kill
```diff
@@ -8382,6 +8393,17 @@ This option makes pt-kill kill matching queries.  This requires MySQL 5.0 or
 newer.  Unlike L<"--kill"> which kills the connection for matching queries,
 this option only kills the query, not its connection.
 
+=item --logdb
```
svetasmirnova, line 8396: Not very intuitive name for the option.

svetasmirnova:
There is such an option already: --log-dsn

PR #435, opened 2020-01-25
    PT-1804  Query information schema for list of tables

ernesto-alikarami:
Modified line 7810 of pt-table-checksum to fetch tables from information schema rather than the actual catalog.

Also added aliases to make sure returned value would be evaluated in the same way.

This will make execution of the script faster as previous query is often slow, depending on the number of tables and usage of the catalog

### DIFF
--- a/bin/pt-table-checksum
+++ b/bin/pt-table-checksum
@@ -7807,7 +7807,8 @@ sub _iterate_dbh {
       }
 
       if ( !$self->{tbls} ) {
-         my $sql = 'SHOW /*!50002 FULL*/ TABLES FROM ' . $q->quote($self->{db});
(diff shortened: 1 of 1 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-toolkit?pullRequest=435) <br/>All committers have signed the CLA.

svetasmirnova (changes requested):
You made changes in the auto-generated code that will be replaced once we run the update-modules utility. Please port your changes into the SchemaIterator package instead.

percona-csalguero:
@svetasmirnova This one seems to be abandoned and it doesn't have tests.
Can we close it?

svetasmirnova:
@percona-csalguero Agree. Also, SHOW TABLES is mapped to the Information Schema query, so performance should be the same in both cases.

PR #437, opened 2020-03-06
Change jemalloc detection

fmbiete:
Simplify jemalloc detection for cases where is not in standard locations or we have multiple versions available in the system. 
pmap is always the source of truth for what libraries are being used.

### DIFF
--- a/bin/pt-mysql-summary
+++ b/bin/pt-mysql-summary
@@ -2264,30 +2264,14 @@ parse_wsrep_provider_options () {
 }
 
 report_jemalloc_enabled() {
-  local JEMALLOC_STATUS=''
-  local GENERAL_JEMALLOC_STATUS=0
-  local JEMALLOC_LOCATION=''
-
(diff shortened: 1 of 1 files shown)


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-toolkit?pullRequest=437) <br/>All committers have signed the CLA.

fmbiete:
This would also solve: PT-1628

svetasmirnova (changes requested):
You made changes in the auto-generated code that will be overridden by the update-modules utility. Please, move your change into the report_mysql_info package. 
Please also add a test case. You will find examples in the t directory.

svetasmirnova:
We cannot repeat PT-1628 anymore. This contribution does not test cases, demonstrating the issue and its resolution, so rejecting it. If you can update it to the current version, put the fix into the proper place, and add test cases, feel free to re-open it.

PR #438, opened 2020-03-25
PT-1810 Update UUID package

percona-csalguero:
Replaced the old UUID package in favor of Google's uuid package for Go


src/go/lib/versioncheck/version_check.go
```diff
@@ -52,11 +52,11 @@ func CheckUpdates(toolName, version string) (string, error) {
 }
 
 func checkUpdates(url string, timeout time.Duration, toolName, version string) (string, error) {
-
 	client := &http.Client{
 		Timeout: timeout,
 	}
-	payload := fmt.Sprintf("%x;%s;%s", uuid.NewV2(uuid.DomainOrg).String(), PERCONA_TOOLKIT, version)
+	vuuid := uuid.New()
```
artemgavrilov: I prefer not to create single-purpose variables and suggest to use `uuid.New().String()` directly inside `Sprintf`, but it's up to you 🙂

PR #446, opened 2020-05-15
PT-1822 Fixed get hostnames for standalone

percona-csalguero:
- [ ] FB: https://github.com/Percona-Lab/pmm-submodules/pull/880

### DIFF
--- a/src/go/mongolib/util/util.go
+++ b/src/go/mongolib/util/util.go
@@ -2,7 +2,6 @@ package util
 
 import (
 	"context"
-	"fmt"
 	"sort"
 	"strings"
 
@@ -11,6 +10,10 @@ import (
 	"go.mongodb.org/mongo-driver/bson/primitive"
 	"go.mongodb.org/mongo-driver/mongo"
 	"go.mongodb.org/mongo-driver/mongo/options"
+)
+
+const (
+	shardingNotEnabledError = 203
 )
 
 var (
@@ -119,6 +122,9 @@ func GetHostnames(ctx context.Context, client *mongo.Client) ([]string, error) {
 	var shardsMap proto.ShardsMap
 	smRes := client.Database("admin").RunCommand(ctx, primitive.M{"getShardMap": 1})
 	if smRes.Err() != nil {
+		if e, ok := smRes.Err().(mongo.CommandError); ok && e.Code == shardingNotEnabledError {
+			return nil, nil // standalone instance
+		}
 		return nil, errors.Wrap(smRes.Err(), "cannot getShardMap for GetHostnames")
 	}
 	if err := smRes.Decode(&shardsMap); err != nil {
@@ -134,7 +140,7 @@ func GetHostnames(ctx context.Context, client *mongo.Client) ([]string, error) {
 		}
 	}
 
-	return nil, fmt.Errorf("cannot get shards map")
+	return nil, nil // standalone instance
 }
 
 func buildHostsListFromReplStatus(replStatus proto.ReplicaSetStatus) []string {
--- a/src/go/mongolib/util/util.go
+++ b/src/go/mongolib/util/util.go
@@ -13,11 +13,12 @@ import (
 )
 
 const (
-	shardingNotEnabledError = 203
+	shardingNotEnabledErrorCode = 203
 )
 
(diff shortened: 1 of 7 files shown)


src/go/mongolib/util/util.go
```diff
@@ -119,6 +122,9 @@ func GetHostnames(ctx context.Context, client *mongo.Client) ([]string, error) {
 	var shardsMap proto.ShardsMap
 	smRes := client.Database("admin").RunCommand(ctx, primitive.M{"getShardMap": 1})
 	if smRes.Err() != nil {
+		if e, ok := smRes.Err().(mongo.CommandError); ok && e.Code == shardingNotEnabledError {
+			return nil, nil // standalone instance
```
BupycHuk, line 126: I would suggest to return error like ErrShardingNotEnabled and handle it where we call this method.

src/go/mongolib/util/util.go
```diff
@@ -134,7 +140,7 @@ func GetHostnames(ctx context.Context, client *mongo.Client) ([]string, error) {
 		}
 	}
 
-	return nil, fmt.Errorf("cannot get shards map")
+	return nil, nil // standalone instance
```
BupycHuk, line 143: when it's possible to not get shardingNotEnabledError on standalone instance?
Is it possible that sharding is enabled, but we got empty `shardsMap`?

src/go/mongolib/util/util_test.go
```diff
@@ -156,6 +161,10 @@ func TestGetShardedHosts(t *testing.T) {
 			if err != nil {
 				t.Errorf("Cannot get a new client for host %s: %s", test.uri, err)
 			}
+			if client == nil {
+				panic(fmt.Sprintf("i: %d, uri: %s\n", i, test.uri))
```
BupycHuk, line 165: it's not a good idea to return panic here.

src/go/mongolib/util/util.go
BupycHuk: @percona-csalguero ^

percona-csalguero: AFAIK, no

src/go/mongolib/util/util.go
```diff
@@ -13,8 +12,13 @@ import (
 	"go.mongodb.org/mongo-driver/mongo/options"
 )
 
+const (
+	shardingNotEnabledErrorCode = 203
+)
+
 var (
-	CANNOT_GET_QUERY_ERROR = errors.New("cannot get query field from the profile document (it is not a map)")
+	CANNOT_GET_QUERY_ERROR  = errors.New("cannot get query field from the profile document (it is not a map)")
```
BupycHuk, line 20: Why do we have this two different styles of naming?

src/go/mongolib/util/util.go
BupycHuk: So why do we have these conditions above?

src/go/mongolib/util/util.go
```diff
@@ -92,7 +96,7 @@ func GetReplicasetMembers(ctx context.Context, clientOptions *options.ClientOpti
 			membersMap[m.Name] = m
 		}
 
-		client.Disconnect(ctx)
+		client.Disconnect(ctx) //nolint
```
BupycHuk, line 99: Looks like we don't disconnect in case we get error.

BupycHuk: https://github.com/percona/percona-toolkit/pull/446/files#diff-a17f19d09995b36b3c78662622d66165R81
https://github.com/percona/percona-toolkit/pull/446/files#diff-a17f19d09995b36b3c78662622d66165R77

BupycHuk: We can extract it to separate task and fix it there.

percona-csalguero: Yes please. It is in a for-loop where a new client is being created for each host found. 
The whole for-loop should be extracted into a new function to make it more clear and it will be possible to use a defer.

percona-csalguero: Created https://jira.percona.com/browse/PT-1855 to make the improvements

PR #454, opened 2020-06-21
MongoDB EXPLAIN JSON APIs

percona-csalguero:
When migrated to the new MongoDB driver, the bson.D data type used in [proto.SystemProfile](https://github.com/percona/percona-toolkit/blob/3.0/src/go/mongolib/proto/system.profile.go) has changed so, the [Query](https://github.com/percona/percona-toolkit/blob/3.0/src/go/mongolib/proto/system.profile.go#L82) field is being marshaled in an incompatible way, adding internal bson references like **$numberLong** to the query.
In this PR, the data files for fingerprints are broken for the new bson.D data type. 
This is WIP - Do not merge.


src/go/mongolib/fingerprinter/fingerprinter.go
```diff
@@ -271,3 +269,80 @@ func deduplicate(s []string) (r []string) {
...
+	// Proper way to detect if protocol used is "op_msg" or "op_command"
+	// would be to look at "doc.Protocol" field,
+	// however MongoDB 3.0 doesn't have that field
+	// so we need to detect protocol by looking at actual data.
+	query := doc.Query
+	if len(doc.Command) > 0 {
+		query = doc.Command
+		if doc.Op == "update" || doc.Op == "remove" {
+			if squery, ok := query.Map()["q"]; ok {
+				switch v := squery.(type) {
```
BupycHuk, line 284: can we use `asMap` here?

src/go/pt-mongodb-summary/oplog/oplog_test.go
```diff
@@ -63,6 +63,7 @@ func TestGetOplogCollection(t *testing.T) {
 }
 
 func TestGetOplogInfo(t *testing.T) {
+	t.Skip("llll")
```
BupycHuk, line 66: Why it's skipped?

src/go/mongolib/fingerprinter/fingerprinter.go
```diff
@@ -120,21 +118,20 @@ func (f *Fingerprinter) Fingerprint(doc proto.SystemProfile) (Fingerprint, error
...
-						if keysMap, ok := f.(bson.M); ok {
-							retKeys = append(retKeys, keys(keysMap, []string{})...)
-						}
-					}
-					if f, ok := m["ns"]; ok {
-						if ns, ok := f.(string); ok {
-							collection = ns
-						}
+				m, err := asMap(g)
+				if err != nil {
```
BupycHuk, line 122: previously we were ignoring in case we can't cast.

BupycHuk: @percona-csalguero ^

src/go/mongolib/fingerprinter/fingerprinter.go
BupycHuk: There no more check if `query.Map()["q"]` exists or not

BupycHuk: @percona-csalguero

percona-csalguero: No. It will return an error if it not exists.

PR #457, opened 2020-07-05
PT-1867 : init recovery tool

AlyHKafoury:


it-percona:
[CLA assistant check](https://cla-assistant.percona.com:443/percona/percona-toolkit?pullRequest=457) <br/>All committers have signed the CLA.

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -136,11 +136,82 @@ func PodZeroReady() error {
...
+		"jsonpath='{.status.phase}'",
+	}
+	output, err := helpers.RunCmd(args...)
+	if err != nil {
+		return false, err
+	}
+	if output == phase {
+		return true, nil
+	}
+	return false, nil
```
dAdAbird:
```go
return output == phase, nil
```

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -136,11 +136,82 @@ func PodZeroReady() error {
...
+	for i := 0; i < clusterSize; i++ {
+		running := false
+		var err error
+		for running == false {
+			time.Sleep(time.Second * 10)
+			running, err = CheckPodPhase(i, "Running")
+			if err != nil {
+				return err
+			}
+		}
```
dAdAbird: should be restricted by time?

src/go/pt-k8s-pxc-recovery/helpers/helpers.go
```diff
@@ -0,0 +1,36 @@
+package helpers
```
dAdAbird: Try to avoid too general packages
https://blog.golang.org/package-names#TOC_5.
https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

src/go/pt-k8s-pxc-recovery/helpers/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"os/exec"
+	"runtime"
+)
+
+var namespace = ""
```
dAdAbird: Try to avoid global states
https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_avoid_package_level_state

src/go/pt-k8s-pxc-recovery/helpers/helpers.go
```diff
@@ -0,0 +1,36 @@
...
+func SetNamespace(name string) {
+	if namespace == "" {
+		namespace = name
+	}
+}
+
+func RunCmd(args ...string) (string, error) {
+	args = append([]string{"--namespace", namespace}, args...)
+	cmd := exec.Command(getKubectl(), args...)
+	println(cmd.String())
```
dAdAbird: 1. use fmt package
2. do we need this? or it's a just debug leftover?

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,217 @@
...
+	return podNames, nil
+}
+
+func ConfirmCrashedStatus() error {
+	podNames, err := getPods()
+	if err != nil {
+		return err
+	}
+	if len(podNames) > 1 {
+		return errors.New("Cluster has more than one Pod up")
```
dAdAbird: https://github.com/golang/go/wiki/CodeReviewComments#error-strings

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,217 @@
...
+	"errors"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/percona/percona-toolkit/src/go/pt-k8s-pxc-recovery/helpers"
+)
+
+var clusterName = ""
+var clusterSize = 0
```
dAdAbird: https://dave.cheney.net/practical-go/presentations/gophercon-singapore-2019.html#_avoid_package_level_state

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,217 @@
...
+	strSize, err := helpers.RunCmd(args...)
+	if err != nil {
+		return err
+	}
+	strSize = "3"
+	clusterSize, err = strconv.Atoi(string(strSize))
+	if err != nil {
+		return err
+	}
+	return nil
```
dAdAbird: 1. just `return err`
2. Add more context to errors. Later on, it's gonna be difficult to figure out where and why an error happened

src/go/pt-k8s-pxc-recovery/helpers/helpers.go
```diff
@@ -0,0 +1,40 @@
...
+	args = append([]string{"-v=0", "--namespace", namespace}, args...)
+	cmd := exec.Command(getKubectl(), args...)
+	cmd.Env = os.Environ()
+	println(cmd.String())
+	stdouterr, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", errors.New(string(stdouterr))
+	}
+	output := string(stdouterr)
+	fmt.Println(output)
```
dAdAbird: do we need this?

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,243 @@
...
+		logs, err := helpers.RunCmd("logs", pod)
+		if err != nil {
+			return err
+		}
+		if strings.Contains(string(logs), "grastate.dat") && strings.Contains(string(logs), "safe_to_bootstrap") {
+			failedNodes++
+		}
+	}
+	if len(podNames) != failedNodes {
+		return errors.New("Not all cluster is down restart failed pods manually")
```
dAdAbird: 1. https://github.com/golang/go/wiki/CodeReviewComments#error-strings
2. Rephrase, please, error message. Now it's a bit confusing

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+		return []string{}, err
+	}
+	formatedOutput := strings.Split(string(out), "\n")
+	podNames := []string{}
+	for _, v := range formatedOutput {
+		podName := strings.Split(v, " ")[0]
+		if strings.Contains(podName, clusterName) && strings.Contains(podName, "pxc") {
+			podNames = append(podNames, podName)
+		}
+	}
```
dAdAbird: I'd rather use k8s lables and metadata:
```
kubectl get pods --no-headers -o custom-columns=":metadata.name" --selector="app.kubernetes.io/component=pxc" --selector="app.kubernetes.io/instance=CLUSTER_NAME"
```

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+func ConfirmCrashedStatus() error {
+	podNames, err := getPods()
+	if err != nil {
+		return err
+	}
+	failedNodes := 0
+	for _, pod := range podNames {
+		logs, err := helpers.RunCmd("logs", pod)
+		if err != nil {
+			return err
```
dAdAbird: here and in other places - add more context to returned errors

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+func RestartPods() error {
+	for i := 0; i < clusterSize; i++ {
+		args := []string{
+			"delete",
+			"pod",
+			clusterName + "-pxc-" + strconv.Itoa(i),
+			"--force",
+			"--grace-period=0",
+		}
+		helpers.RunCmd(args...)
```
dAdAbird: handle error

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+	}
+	return nil
+}
+
+func CheckPodStatus(podID int) (bool, error) {
```
dAdAbird: s/CheckPodStatus/CheckPodReady

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+func PodZeroReady() error {
+	podZeroStatus := false
+	var err error
+	for podZeroStatus != true {
+		time.Sleep(time.Second * 10)
+		podZeroStatus, err = CheckPodStatus(0)
+		if err != nil {
+			return err
+		}
+	}
```
dAdAbird: what if pod never became ready?

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+	if strings.Trim(output, "'") == "true" {
+		return true, nil
+	}
+	return false, nil
+}
+
+func PodZeroReady() error {
+	podZeroStatus := false
+	var err error
+	for podZeroStatus != true {
```
dAdAbird: s/podZeroStatus != true/!podZeroStatus

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+	if strings.Trim(output, "'") == phase {
+		return true, nil
+	}
+	return false, nil
+}
+
+func AllPodsRunning() error {
+	for i := 0; i < clusterSize; i++ {
+		running := false
+		for running == false {
```
dAdAbird: s/running == false/!running

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+			time.Sleep(time.Second * 10)
+			running, _ = CheckPodPhase(i, "Running")
```
dAdAbird: handle error

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+	return nil
+}
```
dAdAbird: 1. If some of the pod(s) didn't reach "Running" status you still return nil and go further. Which is wrong.
2. I don't like this idea of one giant sleep on status checks. It should be smaller pauses but multiple tries in order not to wait unnecessary time. Moreover, 10 sec may not be always enough

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,34 @@
...
+
+	stepOrError(recover.SetClusterSize())
+	stepOrError(recover.ConfirmCrashedStatus())
+	stepOrError(recover.PatchClusterImage())
+	stepOrError(recover.RestartPods())
+	stepOrError(recover.PodZeroReady())
+	stepOrError(recover.AllPodsRunning())
+	stepOrError(recover.SetSSTInProgress())
+	stepOrError(recover.AllPodsReady())
+	stepOrError(recover.FindMostRecentPod())
```
dAdAbird: Add more info. The user should see some progress - on which stage we are now etc. The same with error messages - it should be clear on what stage error happened

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,34 @@
...
+	"flag"
+	"log"
+
+	"github.com/percona/percona-toolkit/src/go/pt-k8s-pxc-recovery/helpers"
+	"github.com/percona/percona-toolkit/src/go/pt-k8s-pxc-recovery/recover"
+)
+
+func stepOrError(err error) {
+	if err != nil {
+		log.Fatal(err)
```
dAdAbird: make obvious for the user that this is an error, e. g. add "Error: " prefix to the output

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+		"pxc",
+		clusterName,
+		"-o",
+		"jsonpath='{.spec.pxc.size}'",
+	}
+	strSize, err := helpers.RunCmd(args...)
+	strSize = strings.Trim(strSize, "'")
+	if err != nil {
+		return err
+	}
```
dAdAbird: The error should be right checked after `RunCmd`

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+	var podID int
+	seqNo := 0
+	for i := 0; i < clusterSize; i++ {
+		output, err := RunCommandInPod(i, "cat", "/var/lib/mysql/grastate.dat")
+		if err != nil {
+			return err
+		}
+		re := regexp.MustCompile(`(?m)seqno:\s*(\d*)`)
+		match := re.FindSubmatch([]byte(output))
+		currentSeqNo, err := strconv.Atoi(string(match[1]))
```
dAdAbird, line 257: it's gonna be panic if len(match) < 2

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,233 @@
...
+		if err != nil {
+			return err
+		}
+		if currentSeqNo > seqNo {
+			seqNo = currentSeqNo
+			podID = i
+		}
+	}
+	mostRecentPod = podID
+	fmt.Println(mostRecentPod)
```
dAdAbird: do we need this?

dAdAbird (changes requested):

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: debug

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: fixed

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: done

src/go/pt-k8s-pxc-recovery/main.go
AlyHKafoury: fixed

src/go/pt-k8s-pxc-recovery/main.go
AlyHKafoury: will do

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: what do you mean here ?

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,230 @@
...
+
+func (c *Cluster) FindMostRecentPod() error {
+	var podID int
+	seqNo := 0
+	for i := 0; i < c.Size; i++ {
+		output, err := c.RunCommandInPod(i, "cat", "/var/lib/mysql/grastate.dat")
+		if err != nil {
+			return err
+		}
+		re := regexp.MustCompile(`(?m)seqno:\s*(\d*)`)
```
dAdAbird: Should we call it on each iteration?

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,230 @@
...
+		"jsonpath='{.spec.pxc.size}'",
+	}
+	strSize, err := kubectl.RunCmd(c.Namespace, args...)
+	if err != nil {
+		return err
+	}
+	strSize = strings.Trim(strSize, "'")
+	c.Size, err = strconv.Atoi(strSize)
+	if err != nil {
+		return errors.New("error getting cluster size, " + err.Error())
```
dAdAbird: use fmt.Errorf

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+func (c *Cluster) FindMostRecentPod() error {
+	var podID int
+	seqNo := 0
+	for i := 0; i < c.Size; i++ {
+		output, err := c.RunCommandInPod(i, "cat", "/var/lib/mysql/grastate.dat")
+		if err != nil {
+			return err
+		}
+		re := regexp.MustCompile(`(?m)seqno:\s*(\d*)`)
+		match := re.FindSubmatch([]byte(output))
```
dAdAbird: You can use [FindStringSubmatch()](https://pkg.go.dev/regexp?tab=doc#Regexp.FindStringSubmatch) so no need in []byte<->string convertion.

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+		"pod",
+		c.Name + "-pxc-" + strconv.Itoa(c.MostRecentPod),
+		"--force",
+		"--grace-period=0",
+	}
+	_, err := kubectl.RunCmd(c.Namespace, args...)
+	if err != nil {
+		return err
+	}
+	return nil
```
dAdAbird: just
```go
return err
```

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,81 @@
...
+	namespace, clusterName := "", ""
+	flag.StringVar(&namespace, "namespace", "default", "Select the namespace in which the cluster is deployed in")
+	flag.StringVar(&clusterName, "cluster", "test-cluster", "Select the cluster to recover")
+	flag.Parse()
+	c := recover.Cluster{Namespace: namespace, Name: clusterName}
+	log.SetPrefix("\n" + log.Prefix())
+
+	log.Printf("Starting recovery process")
+	go func() {
+		for true {
```
dAdAbird:
```suggestion
		for {
```

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,81 @@
...
+	}()
+
+	log.Printf("Getting cluster size")
+	stepOrError(c.SetClusterSize())
+
+	log.Printf("Confirming crashed status")
+	stepOrError(c.ConfirmCrashedStatus())
+
+	log.Printf("Patching cluster image")
+	stepOrError(c.PatchClusterImage("percona/percona-xtradb-cluster:8.0.19-10.1-debug"))
```
dAdAbird: PXC version shouldn't be hardcoded. Or the tool does work only with `8.0.19-10.1` version?

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,81 @@
...
+
+	log.Printf("Finding the most recent pod")
+	stepOrError(c.FindMostRecentPod())
+
+	log.Printf("Recovering most recent pod")
+	go func() {
+		stepOrError(c.RecoverMostRecentPod())
+	}()
+
+	time.Sleep(10 * time.Second)
```
dAdAbird: What does this code should do? Why goroutine? Why 10-sec sleep?

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+func (c *Cluster) ConfirmCrashedStatus() error {
+	podNames, err := c.getPods()
+	if err != nil {
+		return err
+	}
+	failedNodes := 0
+	for _, pod := range podNames {
+		logs, err := kubectl.RunCmd(c.Namespace, "logs", pod)
+		if err != nil {
+			return errors.New("error confirming crashed cluster status" + err.Error())
```
dAdAbird: use fmt.Errorf

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+		if strings.Contains(string(logs), "grastate.dat") && strings.Contains(string(logs), "safe_to_bootstrap") {
+			failedNodes++
+		}
+	}
+	if len(podNames) != failedNodes {
+		return errors.New("found more than one pod running, can't use recovery tool, please manually restart failed pods")
```
dAdAbird:
```suggestion
		return errors.New("found more than one pod running, can't use recovery tool, please restart failed pods manually")
```

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+func (c *Cluster) getPods() ([]string, error) {
+	args := []string{
+		"get",
+		"pods",
+	}
+	out, err := kubectl.RunCmd(c.Namespace, args...)
+	if err != nil {
+		return []string{}, err
+	}
+	formatedOutput := strings.Split(string(out), "\n")
```
dAdAbird: `kubectl.RunCmd` already returns a string. No need in conversion

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+func main() {
+	namespace, clusterName := "", ""
+	flag.StringVar(&namespace, "namespace", "default", "Select the namespace in which the cluster is deployed in")
+	flag.StringVar(&clusterName, "cluster", "test-cluster", "Select the cluster to recover")
+	flag.Parse()
+	c := recover.Cluster{Namespace: namespace, Name: clusterName}
+	log.SetPrefix("\n" + log.Prefix())
```
dAdAbird, line 25: why do you need this?

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+	}
+	c.MostRecentPod = podID
+	return nil
+}
+
+func (c *Cluster) RecoverMostRecentPod() error {
+	_, err := c.RunCommandInPod(c.MostRecentPod, "mysqld", "--wsrep_recover")
+	if err != nil {
+		return err
```
dAdAbird: provide more context for errors (here and in other places)

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+		"jsonpath='{.status.containerStatuses[0].ready}'",
+	}
+	output, err := kubectl.RunCmd(c.Namespace, args...)
+	if err != nil {
+		return false, err
+	}
+	if strings.Trim(output, "'") == "true" {
+		return true, nil
+	}
+	return false, nil
```
dAdAbird:
```suggestion
	return output == "'true'", nil
```

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+		_, err := kubectl.RunCmd(c.Namespace, args...)
+		if err != nil && !strings.ContainsAny(err.Error(), "pods") && !strings.ContainsAny(err.Error(), "not found") {
+			return err
+		}
+	}
+	return nil
+}
+
+func (c *Cluster) CheckPodStatus(podID int) (bool, error) {
```
dAdAbird: The method name is misleading. It should be more like `CheckPodReady`

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+	for i := 0; i < c.Size; i++ {
+		args := []string{
+			"delete",
+			"pod",
+			c.Name + "-pxc-" + strconv.Itoa(i),
+			"--force",
+			"--grace-period=0",
+		}
+		_, err := kubectl.RunCmd(c.Namespace, args...)
+		if err != nil && !strings.ContainsAny(err.Error(), "pods") && !strings.ContainsAny(err.Error(), "not found") {
```
dAdAbird: Read the docs on what `strings.ContainsAny` does. It doesn't search for the substring.

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+		"pxc",
+		c.Name,
+		"--type=merge",
+		`--patch={"spec":{"pxc":{"image":"` + image + `"}}}`,
+	}
+	_, err := kubectl.RunCmd(c.Namespace, args...)
+	if err != nil {
+		return err
+	}
+	return nil
```
dAdAbird:
```suggestion
	return err
```

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+	}
+	return strings.Trim(output, "'") == phase, nil
+}
+
+func (c *Cluster) AllPodsRunning() error {
+	for i := 0; i < c.Size; i++ {
+		running := false
+		for !running {
+			time.Sleep(time.Second * 10)
+			running, _ = c.CheckPodPhase(i, "Running")
```
dAdAbird: An error should be handled

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+		return "", err
+	}
+	return output, nil
+}
+
+func (c *Cluster) SetSSTInProgress() error {
+	for i := 0; i < c.Size; i++ {
+		_, err := c.RunCommandInPod(i, "touch", "/var/lib/mysql/sst_in_progress")
+		if err != nil {
+			return err
```
dAdAbird: provide more context for the error. e.g. it would be nice to know on what pod error happened
the same in many places with errors

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,285 @@
...
+		}
+	}
+	return nil
+}
+
+func (c *Cluster) AllPodsReady() error {
+	for i := 0; i < c.Size; i++ {
+		podReadyStatus := false
+		var err error
+		for podReadyStatus == false {
```
dAdAbird:
```suggestion
		for !podReadyStatus {
```

dAdAbird (changes requested):

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,81 @@
...
+
+	log.Printf("Recovering most recent pod")
+	go func() {
+		stepOrError(c.RecoverMostRecentPod())
+	}()
+
+	time.Sleep(10 * time.Second)
+
+	log.Printf("Patching cluster image")
+	stepOrError(c.PatchClusterImage("percona/percona-xtradb-cluster:8.0.19-10.1"))
```
hors: why do we need to hardcode it? the images can be different....

hors (changes requested):

hors (changes requested):
Is it ok that we have the cluster in the following state:
```
cluster1-pxc-0                                     0/1     Running   0          25s
cluster1-pxc-1                                     1/1     Running   0          87s
cluster1-pxc-2                                     0/1     Running   0          40s
```
but we reported that 'Completed the restore process'? Maybe we need to wait until all pods are ready...

In a few minutes I have: 
```
cluster1-pxc-0                                     1/1     Running            0          8m18s
cluster1-pxc-1                                     0/1     CrashLoopBackOff   4          9m20s
cluster1-pxc-2                                     0/1     CrashLoopBackOff   5          8m33s

```
so recovery  tool does not work for me.
@AlyHKafoury  please check it. 

P.S. I have checked it several times.

hors (changes requested):
Hi @AlyHKafoury 
I have cluster that was ok but this check passed, so maybe we need to check that cluster is really crashed e.g. check statuses of the pods...
`2020/08/21 17:02:05 Confirming crashed status`

Now we have infinity loop in `2020/08/21 15:32:47 Make sure pod zero is ready
` I think we need to add timeouts for all steps @delgod what do you think?

AlyHKafoury:
> Hi @AlyHKafoury
> I have cluster that was ok but this check passed, so maybe we need to check that cluster is really crashed e.g. check statuses of the pods...
> `2020/08/21 17:02:05 Confirming crashed status`
> 
> Now we have infinity loop in `2020/08/21 15:32:47 Make sure pod zero is ready ` I think we need to add timeouts for all steps @delgod what do you think?

I will re check it with an ok cluster

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: fixed

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: fixed

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: fixed

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: fixed

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: fixed

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: fixed

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: fixed

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: these errors are kubectl error messages they are pretty verbose

src/go/pt-k8s-pxc-recovery/main.go
AlyHKafoury: yes this is needed to help format the output correctly

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: fixed

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: same these kubectl errors they are pretty verbose

src/go/pt-k8s-pxc-recovery/main.go
AlyHKafoury: This code runes some steps in parrellel with the other steps which is needed for the recovery process and the 10 seconds is some time needed to give it a head start it needs

src/go/pt-k8s-pxc-recovery/recover/recover.go
dAdAbird: 1. You're running a kubectl command via exec.Cmd. So error could happen out of the scope of kubectl
2. Some extra context won't make any worse. This tool is intended to be run by customers. If something went wrong we should have as much info as we can to tackle the problem. And we won't have a chance to adjust error messaging in the tool and ask to rerun it so we could understand e.g. on what pod error happened. 

And this is related not only to this code but to error handling in the tool in general. Wrap errors with some context info before return!!!

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+	log.Printf("Recovering most recent pod")
+	go func() {
+		err := c.RecoverMostRecentPod()
+		if err != nil {
+			log.Printf("Recovering most recent pod still in progress")
+		}
+	}()
+
+	time.Sleep(10 * time.Second)
```
dAdAbird, line 74: Why 10 sec?  Are you sure 10sec. gonna work for any env/cluster etc.?
What if `"mysqld", "--wsrep_recover"` will take more than 10 sec?
I reckon we shouldn't just run `mysqld` at the end of this func, but wait until "_The mysqld process will initialize the database once again, and it will be available for the incoming connections._"

You shouldn't rely on any sleep in this situation

src/go/pt-k8s-pxc-recovery/main.go
dAdAbird: but what is incorrect without an extra "\n"? We should have blank lines between log messages?
Could you please show an example of the correct and incorrect output?

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,297 @@
...
+		"pod",
+		c.Name + "-pxc-0",
+		"-o",
+		"jsonpath='{.spec.containers[0].image}'",
+	}
+	clusterImage, err := kubectl.RunCmd(c.Namespace, args...)
+	if err != nil {
+		return err
+	}
+	c.ClusterImage = strings.Trim(clusterImage, "'")
```
dAdAbird: I don't see any usage of `ClusterImage` later on

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,297 @@
...
+		"jsonpath='{.spec.pxc.size}'",
+	}
+	strSize, err := kubectl.RunCmd(c.Namespace, args...)
+	if err != nil {
+		return err
+	}
+	strSize = strings.Trim(strSize, "'")
+	c.Size, err = strconv.Atoi(strSize)
+	if err != nil {
+		return fmt.Errorf("error getting cluster size, %s", err.Error())
```
dAdAbird:
```suggestion
		return fmt.Errorf("error getting cluster size, %v", err)
```

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,297 @@
...
+	formatedOutput := strings.Split(out, "\n")
+	podNames := []string{}
+	for _, v := range formatedOutput {
+		podName := strings.Split(v, " ")[0]
+		if strings.Contains(podName, c.Name) && strings.Contains(podName, "pxc") {
+			podNames = append(podNames, podName)
+		}
+	}
+	return podNames, nil
+}
```
dAdAbird: I suggest using `kubectl get pods --no-headers -o custom-columns=":metadata.name" --selector=...`
With an appropriate selector you'll be able to get only pxc pods. Plus `-o custom-columns=` returns only pod names so no need in extra parsing. It seems much more reliable.

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,297 @@
...
+		}
+		if strings.Contains(logs, "grastate.dat") && strings.Contains(logs, "safe_to_bootstrap") {
+			failedNodes++
+		}
+	}
+	if len(podNames) != failedNodes {
+		return fmt.Errorf("found more than one pod running, can't use recovery tool, please restart failed pods manually")
+	}
+	return nil
+}
```
dAdAbird: What this code intends to do? _I see what it does but I'm interested in what **should** be done_
The predicate `len(podNames) != failedNodes`  and error "found more than one pod running..." are confusing. Why "more than one"? Is it ok to have one pod not crashed and running? If so - this code doesn't allow this

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,297 @@
...
+	seqNo := 0
+	re := regexp.MustCompile(`(?m)seqno:\s*(\d*)`)
+	for i := 0; i < c.Size; i++ {
+		output, err := c.RunCommandInPod(i, "cat", "/var/lib/mysql/grastate.dat")
+		if err != nil {
+			return err
+		}
+		match := re.FindStringSubmatch(output)
+		if len(match) < 2 {
+			return fmt.Errorf("unable to get seqno")
```
dAdAbird: This is a good example of the necessity to provide more context to the errors. Imagine error here. A user would see
```
...
Finding the most recent pod
Error: unable to get seqno
```
how useful is information in this error?

what the heck is seqno? on what pod it happend? why did it happen?

dAdAbird (changes requested):

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,297 @@
...
+	_, err := kubectl.RunCmd(c.Namespace, args...)
+	return err
+}
+
+func (c *Cluster) RestartPods() error {
+	for i := 0; i < c.Size; i++ {
+		args := []string{
+			"delete",
+			"pod",
+			c.Name + "-pxc-" + strconv.Itoa(i),
```
dAdAbird: Here and everywhere else. You shouldn't stick to the pod naming conventions (`namespace`+`-pxc-`+`podNo`) but rather  get pxc pods list by appropriate selectors and than use it

dAdAbird (changes requested):

src/go/pt-k8s-pxc-recovery/main.go
dAdAbird: Ah, ok, I got it. It is because of dots printing. Nice.

src/go/pt-k8s-pxc-recovery/main.go
delgod: this process on high-loaded database can take a while

delgod (changes requested):

src/go/pt-k8s-pxc-recovery/main.go
AlyHKafoury: @dAdAbird the 10 seconds is just to give a head start the other features will wait for the mysqld to start

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: @dAdAbird it is used in main line 76

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+	log.Printf("Patching cluster image")
+	stepOrError(c.PatchClusterImage(c.ClusterImage))
```
AlyHKafoury: @dAdAbird

src/go/pt-k8s-pxc-recovery/recover/recover.go
AlyHKafoury: it is supposed to make sure that all pods that are up are in crashed status and if not the recovery tool is not needed you can just restart the failed pods and it will re join the cluster

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,91 @@
...
+
+	log.Printf("Restart all pods execpt most recent pod")
+	stepOrError(c.RestartAllPodsExceptMostRecent())
+
+	log.Printf("Make sure all pods are running")
+	stepOrError(c.AllPodsRunning())
```
hors, line 83: What will we do if mysqld never starts to work correctly? Maybe we need to add some smart logic e.g. add some timeout and periodically check whether SST/IST is in progress or not in order to be sure that recovery process does not hang.

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,302 @@
...
+	if err != nil {
+		return fmt.Errorf("Error getting pods : %s", err)
+	}
+	failedNodes := 0
+	for _, pod := range podNames {
+		logs, err := kubectl.RunCmd(c.Namespace, "logs", pod)
+		if err != nil {
+			return fmt.Errorf("error confirming crashed cluster status %s", err)
+		}
+		if strings.Contains(logs, "grastate.dat") && strings.Contains(logs, "safe_to_bootstrap") {
```
hors: In my case all my clusters passed this check because in all logs we have "grastate.dat" and "safe_to_bootstrap" in different places . It is not correct to check these two different words. You need to check  e.g. 'grastate.dat file manually and set safe_to_bootstrap to 1' this one.

src/go/pt-k8s-pxc-recovery/main.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}()
+
+	time.Sleep(10 * time.Second)
+
+	log.Printf("Patching cluster image")
+	stepOrError(c.PatchClusterImage(c.ClusterImage))
+
+	log.Printf("Restart all pods execpt most recent pod")
+	stepOrError(c.RestartAllPodsExceptMostRecent())
```
hors, line 80: in point 7 in doc we have `rm /var/lib/mysql/sst_in_progress` but you do not do it in this tool. Why?

hors (changes requested):

src/go/pt-k8s-pxc-recovery/main.go
dAdAbird: 1. I still don't understand why 10 sec needed. If we are waiting later why does it needs at all? Why 10? How can you be sure that 10sec is enough for a spinup?
2. I don't see how is the other features is waiting for the mysqld to start. Neither `c.PatchClusterImage` nor `c.RestartAllPodsExceptMostRecent` doesn't wait for anything - it just runs `kubectl`.
3. Suppose something is waiting. Then what happens if `c.RecoverMostRecentPod` fails? Waiting forever?

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,302 @@
...
+	}
+	strSize = strings.Trim(strSize, "'")
+	c.Size, err = strconv.Atoi(strSize)
+	if err != nil {
+		return fmt.Errorf("error getting cluster size, %s", err)
+	}
+	return nil
+}
+
+func (c *Cluster) GetClusterImage() error {
```
dAdAbird: GetClusterImage() sounds like it should **return** an image, but in fact, it **sets** the image

src/go/pt-k8s-pxc-recovery/recover/recover.go
dAdAbird: That's unexpected. If you set ClusterImage here as the part of the `*Cluster` object why it needs to be extracted from this object by the user (_of the package_) and passed as an argument to the method of this object.
Do either
```go
image, err := c.GetClusterImage()
...
c.PatchClusterImage(image)
```
or 
```go
c.SetClusterImage()
c.PatchClusterImage()

...

func (c *Cluster) PatchClusterImage() error {
    // use c.ClusterImage
...
}
```

the current implementation is too confusing

src/go/pt-k8s-pxc-recovery/recover/recover.go
dAdAbird: I can't see any changes or conversation following this comment

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,302 @@
...
+		}
+	}
+	c.MostRecentPod = podID
+	return nil
+}
+
+func (c *Cluster) RecoverMostRecentPod() error {
+	_, err := c.RunCommandInPod(c.MostRecentPod, "mysqld", "--wsrep_recover")
+	if err != nil {
+		return fmt.Errorf("Error recovering most recent pod: %s", err)
```
dAdAbird: 1. Error strings should nor be capitalized. See. https://github.com/golang/go/wiki/CodeReviewComments#error-strings

2. I requested adding a more context not in the sake of more text. But for easier debug should error happen. Adding the same prefix to all errors in the method doesn't help at all. Ok, you may do add the prefix but on the level higher. E.g.
```go
err := c. RecoverMostRecentPod()
if err != nil {
    return fmt.Errorf("error recovering most recent pod: %s", err)
}
```

Here, in the method, make sense to add more context so later it would be easier to identify on what stage error happed. Was it `"mysqld", "--wsrep_recover"` or `grastate patching` etc.?
Even if the method has one error path, yet it could be useful to add more context to provide more info for the user/support/etc. what and why exactly happened without the necessity in source code excavation.

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,302 @@
...
+	return nil
+}
+
+func (c *Cluster) getPods() ([]string, error) {
+	args := []string{
+		"get",
+		"pods",
+		"--no-headers",
+		"-o",
+		"custom-columns=:metadata.name",
```
dAdAbird, line 62: My previous suggestion was more about `--selector=...` usage

> With an appropriate selector you'll be able to get only pxc pods. Plus -o custom-columns= returns only pod names so no need in extra parsing. It seems much more reliable.`

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,302 @@
...
+			podNames = append(podNames, podName)
+		}
+	}
+	return podNames, nil
+}
+
+func (c *Cluster) ConfirmCrashedStatus() error {
+	podNames, err := c.getPods()
+	if err != nil {
+		return fmt.Errorf("Error getting pods : %s", err)
```
dAdAbird, line 81: 1. Here and in many places: Error strings should nor be capitalized. See. https://github.com/golang/go/wiki/CodeReviewComments#error-strings 

2. Here and in many places:  s/"pods :"/"pods:" - there is shouldn't be whitespace before the colon

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,302 @@
...
+		logs, err := kubectl.RunCmd(c.Namespace, "logs", pod)
+		if err != nil {
+			return fmt.Errorf("error confirming crashed cluster status %s", err)
+		}
+		if strings.Contains(logs, "grastate.dat") && strings.Contains(logs, "safe_to_bootstrap") {
+			failedNodes++
+		}
+	}
+	if len(podNames) != failedNodes {
+		return fmt.Errorf("found more than one pod running, can't use recovery tool, please restart failed pods manually")
```
dAdAbird: This error message is confusing. More than one means 2+. But as I understand from your previous comment and as I see from code **_all_**  of the running pods should be in a crushed state. So the message should rather state "one or more ..."

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,302 @@
...
+	for _, pod := range podNames {
+		logs, err := kubectl.RunCmd(c.Namespace, "logs", pod)
+		if err != nil {
+			return fmt.Errorf("error confirming crashed cluster status %s", err)
+		}
+		if strings.Contains(logs, "grastate.dat") && strings.Contains(logs, "safe_to_bootstrap") {
+			failedNodes++
+		}
+	}
+	if len(podNames) != failedNodes {
```
dAdAbird: "`all pods should be in a crashed state`" == "`one not_crused pod -> error`"
So the nominal algorithm here should be simplified to:
```go
for  _, pod := range pods {
    if !isCrashed(pod) {
        return errors.New("one or more pod ...")
    }
}
```

no need in `failedNodes` counter and later `if len(podNames) != failedNodes`

dAdAbird (changes requested):

src/go/pt-k8s-pxc-recovery/main.go
dAdAbird: @AlyHKafoury I see no changes on that matter

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,326 @@
...
+	out, err := kubectl.RunCmd(c.Namespace, args...)
+	if err != nil {
+		return []string{}, err
+	}
+	formatedOutput := strings.Split(out, "\n")
+	podNames := []string{}
+	for _, podName := range formatedOutput {
+		if strings.Contains(podName, c.Name) && strings.Contains(podName, "pxc") {
+			podNames = append(podNames, podName)
+		}
```
dAdAbird, line 73: As it mentioned here https://github.com/percona/percona-toolkit/pull/457/files#r489615730? please use selectors (https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) to get pods. Strings sub matching isn't so reliable. All pxc nodes have labels, use it https://github.com/percona/percona-xtradb-cluster-operator/blob/master/pkg/pxc/app/statefulset/node.go#L37-L43

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,326 @@
...
+	if err != nil {
+		return false, fmt.Errorf("error checking pod ready: %s", err)
+	}
+	return strings.Trim(output, "'") == "true", nil
+}
+
+func (c *Cluster) PodZeroReady() error {
+	podNames, err := c.getPods()
+	if err != nil {
+		return err
```
dAdAbird, line 147: here and in all new code - please, wrap errors!!!

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,326 @@
...
+	}
+	podZeroStatus := false
+	for !podZeroStatus {
+		time.Sleep(time.Second * 10)
+		podZeroStatus, err = c.CheckPodReady(podNames[0])
```
dAdAbird, line 152: I wouldn't rely on kubectl sorting. Here is the place I'd check the suffix in podName to be sure

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,326 @@
...
+		if err != nil {
+			return err
+		}
+	}
```
dAdAbird, line 156: If pod never gets ready application will be stuck forever. Here (and in other places witch infinite loop) should be some deadline and handling of the situations when the deadline is reached.

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,326 @@
...
+
+func (c *Cluster) SetSSTInProgress() error {
+	podNames, err := c.getPods()
+	if err != nil {
+		return err
+	}
+	for _, podName := range podNames {
+		_, err := c.RunCommandInPod(podName, "touch", "/var/lib/mysql/sst_in_progress")
+		if err != nil {
+			return fmt.Errorf("error setting sst in progress", err)
```
dAdAbird, line 216:
```suggestion
			return fmt.Errorf("error setting sst in progress: %v", err)
```

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,326 @@
...
+			args := []string{
+				"delete",
+				"pod",
+				podName,
+				"--force",
+				"--grace-period=0",
+			}
+			_, err := kubectl.RunCmd(c.Namespace, args...)
+			if err != nil {
+				return fmt.Errorf("error restarting pods : %s", err)
```
dAdAbird, line 306: it would be useful to know on which pod error happened

dAdAbird (changes requested):

src/go/pt-k8s-pxc-recovery/recover/recover.go
```diff
@@ -0,0 +1,326 @@
...
+		return fmt.Errorf("Error getting pods : %s", err)
+	}
+	for _, pod := range podNames {
+		logs, err := kubectl.RunCmd(c.Namespace, "logs", pod)
+		if err != nil {
+			return fmt.Errorf("error confirming crashed cluster status %s", err)
+		}
+		if !strings.Contains(logs, "grastate.dat") && !strings.Contains(logs, "safe_to_bootstrap") &&
+			!strings.Contains(logs, "It may not be safe to bootstrap the cluster from this node") {
+			return fmt.Errorf("found one or more pods in healthy state, can't use recovery tool, please restart failed pods manually")
```
hors, line 90: Please see point 4 in doc https://www.percona.com/doc/kubernetes-operator-for-pxc/recovery.html

We can not say the user "please do something manually" for the tool which should automatically  recover the cluster.
So, you need automatically kill the 'mysqld-ps' process if it is running.

delgod:
automatic recovery was integrated into PXC 1.7.0

PR #459, opened 2020-07-15
PMM-6256 Updated ExplainCmd for MongoDB

percona-csalguero:
In order to make explain for MongoDB work, we need to remove the "$db" field from the explain command generated in proto.system.profile since it is a duplicated field that triggers a MongoDB error.

### DIFF
--- a/src/go/mongolib/proto/system.profile.go
+++ b/src/go/mongolib/proto/system.profile.go
@@ -178,7 +178,6 @@ func (self ExampleQuery) ExplainCmd() bson.D {
 					} else {
 						cmd = append(cmd[:i], cmd[i+1:]...)
 					}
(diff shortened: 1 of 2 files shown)


src/go/mongolib/proto/system.profile.go
```diff
@@ -178,7 +178,7 @@ func (self ExampleQuery) ExplainCmd() bson.D {
 					} else {
 						cmd = append(cmd[:i], cmd[i+1:]...)
 					}
-					break
+					// break LOOP
```
AlekSi, line 181: Should we keep that commented-out?

percona-csalguero: Removed

PR #471, opened 2020-11-05
Support adding explain output for slow query in JSON report as well

perfectayush:
I am generating JSON reports from slow logs and consuming them for a visual dashboard in elasticsearch.
I find it useful to add the EXPLAIN output in the JSON reports as well.


svetasmirnova (changes requested):
You changed the auto-generated part of the bin/pt-query-digest code. These changes will be overwritten once we rebuild the tool using the update-modules utility. Please make changes in the package JSONReportFormatter instead.

Please, also add the test case. You will find examples in the directory t.

svetasmirnova (approved):

EvgeniyPatlan (approved):

PR #474, opened 2020-12-17
Added mysql version check for bug fix introduced in 8.0.14

duxthefux:
According to https://bugs.mysql.com/bug.php?id=89441 this bug has been fixed so we can whitelist everything above 8.0.14.

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -10056,7 +10056,7 @@ sub main {
       }
    }
 
-   if ($vp->cmp('8.0') > -1 && $vp->flavor() !~ m/maria/i && $alter_fk_method eq 'drop_swap') {
+   if (($vp->cmp('8.0') >= 0 && $vp->cmp('8.0.14') <= 0) && $vp->flavor() !~ m/maria/i && $alter_fk_method eq 'drop_swap') {
        my $msg = "--alter-foreign-keys-method=drop_swap doesn't work with MySQL 8.0+\n".
                  "See https://bugs.mysql.com/bug.php?id=89441";
        _die($msg, INVALID_PARAMETERS);


bin/pt-online-schema-change
```diff
@@ -10056,7 +10056,7 @@ sub main {
       }
    }
 
-   if ($vp->cmp('8.0') > -1 && $vp->flavor() !~ m/maria/i && $alter_fk_method eq 'drop_swap') {
+   if (($vp->cmp('8.0') >= 0 && $vp->cmp('8.0.14') <= 0) && $vp->flavor() !~ m/maria/i && $alter_fk_method eq 'drop_swap') {
```
duxthefux, line 10059: The goal is here to only throw an error between 8.0 and 8.0.14. 

I‘m not sure if the usage of cmp() is correct.

ddeschenes-1: seems good to me, although if 8.0.14 fixed it, maybe it should be 
cmp('8.0.14') < 0

ddeschenes-1 (approved):

bin/pt-online-schema-change
duxthefux: shall I change it?

ddeschenes-1: I don't think 1 version off is so bad. Kindof a  little margin? lol

duxthefux: yeah it's one version off into the right direction (8.0.15) and not 8.0.13 right?

ddeschenes-1: I infer so.

benlangfeld:
This fixes https://jira.percona.com/browse/PT-1940. Can it be merged and released please?

percona-csalguero (approved):

PR #478, opened 2021-01-08
[pt-deadlock-logger] add --disable-buffering-stdout option

kenju:
## Problem

It is a common behaviour that Perl uses buffering for I/O
to reduce the number of syscalls from the point of performance.

This can be a problem, for example, when `pt-deadlock-logger` is
running as a Docker container (which means that no terminal is attached)
and the number of deadlock events are scarce.

The `print()`'s output does not appear until the buffer gets filled up,
which prevents users from getting deadlock events as soon as possible.

## Solution

Add `--disable-buffering-stdout` to disable the buffering behaviour.

The buffering should not be disabled by default, because the buffering
is basically for improving the performances by optimizing the number of
syscalls for I/O. That's why the option should be enabled explicitly.

## Further Reading

- https://perl.plover.com/FAQs/Buffering.html

## NOTE

Double-checked that this fixes our problem on the production environment.

### DIFF
--- a/bin/pt-deadlock-logger
+++ b/bin/pt-deadlock-logger
@@ -4650,6 +4650,15 @@ sub main {
    $o->usage_or_errors();
 
    # ########################################################################
+   # Disable buffering log messages to STDOUT
+   # ########################################################################
+   if ( $o->get('disable-buffering-stdout') ) {
+      PTDEBUG && _d('Disabled buffering log messages to STDOUT');
+      use IO::Handle;
+      STDOUT->autoflush(1);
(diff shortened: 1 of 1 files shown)


kenju:
@percona-csalguero Hi :wave:

Would I get a review on this PR, please? Please guide me if I'm missing some contributing rules. 
Thank you in advance.

thom-vend:
Hi @kenju 
I had some buffering issue today, found your PR, seems good, might add for stderr as well!
Why did you close it? wasn't working?

PR #498, opened 2021-07-16
PT-1930: pt-k8s-debug-collector should not attempt collecting PXC information when running against PSMDB

pooknull:
https://jira.percona.com/browse/PT-1930

Removed PXC resources to collect while running with `--resource psmdb`

### DIFF
--- a/src/go/pt-k8s-debug-collector/dumper/dumper.go
+++ b/src/go/pt-k8s-debug-collector/dumper/dumper.go
@@ -61,6 +61,12 @@ func New(location, namespace, resource string) Dumper {
 				"perconaxtradbclusterbackups",
 				"perconaxtradbclusterrestores",
 				"perconaxtradbclusters")
+		} else if resource == "psmdb" {
+			resources = append(resources,
+				"perconaservermongodbbackups",
+				"perconaservermongodbrestores",
+				"perconaservermongodbs",
+			)
 		}
 	}
 	return Dumper{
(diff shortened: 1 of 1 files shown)


src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -58,6 +54,14 @@ func New(location, namespace, resource string) Dumper {
...
 	if len(resource) > 0 {
 		resources = append(resources, resource)
+
+		if resource == "pxc" {
+			resources = append(resources,
+				"perconaxtradbbackups",
+				"perconaxtradbclusterbackups",
+				"perconaxtradbclusterrestores",
+				"perconaxtradbclusters")
+		}
```
delgod, line 69: can we also improve psmdb case?
```suggestion
		} else if resource == "psmdb" {
			resources = append(resources,
				"perconaservermongodbbackups",
				"perconaservermongodbrestores",
				"perconaservermongodbs")
		}
```

delgod (changes requested):

src/go/pt-k8s-debug-collector/dumper/dumper.go
pooknull: I added it in 010c9796a8566039013b9a27d9fc7dcc73e33faf

Also, I think we should use `resourceType` function here from #500
https://github.com/percona/percona-toolkit/blob/c8b821219638c4b819116731dc45592a81556af1/src/go/pt-k8s-debug-collector/dumper/dumper.go#L357

delgod (approved):

qJkee (approved):

percona-csalguero (approved):

PR #499, opened 2021-07-19
Make build reproducible

jelly:
By setting SOURCE_DATE_EPOCH builds the build binaries can become
reproducible as a rebuild will take the old build date as input.

Motivation: https://reproducible-builds.org


jelly:
Here is the diffoscope output of a rebuild showing the timestamp as difference: https://reproducible.archlinux.org/api/v0/builds/131169/diffoscope

jelly:
Ping?

percona-csalguero:
Thanks for this patch.
I've created a Jira to track this issue: https://jira.percona.com/browse/PT-2011

PR #500, opened 2021-07-20
PT-1931: pt-k8s-debug-collector does not gather summaries for PSMDB

pooknull:
https://jira.percona.com/browse/PT-1931

When I tried to reproduce the bug, `pt-mongodb-summary` didn't launch only when option `--cluster` was specified. This PR fixes this behavior and also fixes https://jira.percona.com/browse/PT-1932

### DIFF
--- a/src/go/pt-k8s-debug-collector/dumper/dumper.go
+++ b/src/go/pt-k8s-debug-collector/dumper/dumper.go
@@ -166,12 +166,12 @@ func (d *Dumper) DumpCluster() error {
 				continue
 			}
 			location = filepath.Join(d.location, ns.Name, pod.Name, "/pt-summary.txt")
-			component := d.crType
-			if d.crType == "psmdb" {
+			component := resourceType(d.crType)
+			if component == "psmdb" {
 				component = "mongod"
 			}
 			if pod.Labels["app.kubernetes.io/component"] == component {
-				output, err = d.getPodSummary(d.crType, pod.Name, pod.Labels["app.kubernetes.io/instance"], tw)
+				output, err = d.getPodSummary(resourceType(d.crType), pod.Name, pod.Labels["app.kubernetes.io/instance"], tw)
 				if err != nil {
 					d.logError(err.Error(), d.crType, pod.Name)
 					err = addToArchive(location, d.mode, []byte(err.Error()), tw)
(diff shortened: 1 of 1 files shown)


src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -166,12 +166,12 @@ func (d *Dumper) DumpCluster() error {
...
 			location = filepath.Join(d.location, ns.Name, pod.Name, "/pt-summary.txt")
-			component := d.crType
-			if d.crType == "psmdb" {
+			component := resourceType(d.crType)
+			if component == "psmdb" {
 				component = "mongod"
 			}
 			if pod.Labels["app.kubernetes.io/component"] == component {
-				output, err = d.getPodSummary(d.crType, pod.Name, pod.Labels["app.kubernetes.io/instance"], tw)
+				output, err = d.getPodSummary(resourceType(d.crType), pod.Name, pod.Labels["app.kubernetes.io/instance"], tw)
```
delgod, line 174: feel free to ignore
```suggestion
				output, err = d.getPodSummary(component, pod.Name, pod.Labels["app.kubernetes.io/instance"], tw)
```

delgod (approved):

src/go/pt-k8s-debug-collector/dumper/dumper.go
pooknull: pt-mongodb-summary wouldn't run in this case. `getPodSummary` method expects the first argument to be either `pxc` or `psmdb`. While running against PSMDB, the `component` variable will have `mongod` value

delgod: in such a case why do we expect `psmdb` value in line 170 from the `component` variable?

delgod (changes requested):

delgod (approved):

qJkee (approved):

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -166,12 +166,12 @@ func (d *Dumper) DumpCluster() error {
 				continue
 			}
 			location = filepath.Join(d.location, ns.Name, pod.Name, "/pt-summary.txt")
-			component := d.crType
-			if d.crType == "psmdb" {
+			component := resourceType(d.crType)
+			if component == "psmdb" {
 				component = "mongod"
 			}
```
percona-csalguero, line 172: Variable name is `component` but the function is returning a resource type and that variable is being used only once.
We could replace it with a short if statement.

```suggestion
```
```suggestion
            if rt:= resourceType(d.crType); rt == "psmdb" {
                component = "mongod"
            }

```

percona-csalguero (approved):

PR #508, opened 2021-08-16
Update installation.rst

spron-in:
1. the list of operation systems is outdated
2. not on any installation page we have the list of supported OSes
3. we have supported OSes in our lifecycle doc - why add it here?

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -4,19 +4,6 @@
 Installing Percona Toolkit
 ==========================
 
-Percona provides packages for most popular 64-bit Linux distributions:
-
-* Debian 9 ("Stretch")
-* Debian 10 ("Buster")
-* Ubuntu 18.04 LTS ("Bionic Beaver")
-* Ubuntu 20.04 LTS ("Focal Fossa")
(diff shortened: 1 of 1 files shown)


docs/installation.rst
```diff
@@ -4,19 +4,6 @@
 Installing Percona Toolkit
 ==========================
```
rnovikovP, line 6:
```suggestion

Percona provides binaries and packages for systems listed in the `official support policy <https://www.percona.com/services/policies/percona-software-support-lifecycle#pt>`_.
```

rnovikovP (approved):

rnovikovP:
If we don't have common "text" for referencing Support Policy - let's make it and use it everywhere.

@percona-csalguero , is this file a correct place to add this text to be on the website?

docs/installation.rst
ghost: Converted to rst link format

ghost: Maybe we need to emphasise ...
> Percona provides binaries and packages *only* for systems listed...

ghost (approved):

PR #509, opened 2021-08-17
added new hook before_die

migliozziilaria:
before die, the script calls the new hook (before_die) in oder to execute extra user's operations.

Alternatively, you could call the before_exit hook.


ynuyasha:
Hi @percona-csalguero someone can take care of this PR?

percona-csalguero:
Hi, thanks for the contribution.
Would it be possible to write a test? There are examples in the `t` directory.

Thanks

svetasmirnova (changes requested):
Please add a test case as was suggested by @percona-csalguero

bin/pt-online-schema-change
```diff
@@ -8412,7 +8412,7 @@ my $term_readkey = eval {
 
 use sigtrap 'handler', \&sig_int, 'normal-signals';
 
-
+my $plugin;
```
svetasmirnova, line 8484: Why did you make this variable global? It is enough to have it defined inside the main function.

svetasmirnova: Gotcha: to use it in the sub _die.

svetasmirnova (approved):

EvgeniyPatlan (approved):

PR #516, opened 2021-09-24
PT-1860 make pt-osc respect case insesitive lookup on Windows and osx

ikruglov:
This PR teaches pt-osc how to perform case-insensitive lookup on platform which use case insensitive filesystems, such Windows or osx. Before searching the table via `SHOW TABLES FROM`, pt-osc checks `lower_case_table_names` param and  adjust it string comparison accordingly if `lower_case_table_names > 0`.

According to https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html MySQL 5.6/5.7/8.0 are use case-sensitive search when `lower_case_table_names == 0` and case-insensitive lookup when value is either `1` or `2`.

Before the patch, pt-osc used to fail with `Table does not exist error`.


svetasmirnova (approved):
Please sign CLA, so we can accept your patch.

ikruglov:
Hi @svetasmirnova. I signed CLA. Please check it out. Hope that we can merge this PR.

percona-csalguero (approved):

percona-csalguero:
Hi @ikruglov .
The PR was made with the user `ikruglov-db`. You need to sing the CLA with that user. 
Thanks

bin/pt-online-schema-change
```diff
@@ -3468,6 +3468,22 @@ sub check_table {
...
+   my $lctn_sql = 'SELECT @@lower_case_table_names';
+   PTDEBUG && _d($lctn_sql);
+   my ($lower_case_table_names) = $dbh->selectrow_array($lctn_sql);
+   PTDEBUG && _d("lower_case_table_names=$lower_case_table_names");
+
+   if ($lower_case_table_names > 0) {
+       PTDEBUG && _d("MySQL uses case-insensitive lookup, converting '$tbl' to lowercase");
+       $tbl = lc $tbl;
+   }
+
```
svetasmirnova: This code is in the TableParser package and should be defined in lib/TableParser.pm Otherwise, it will be overwritten once we update the modules. Please adjust your PR.

ikruglov: @svetasmirnova, I've updated the TableParse. Please check it out

svetasmirnova (approved):

EvgeniyPatlan (approved):

PR #517, opened 2021-09-30
PT-1398 Updated way of getting MySQL PID

percona-csalguero:


JiriCtvrtka (approved):
just remove extra spaces and tabs in some places

PR #518, opened 2021-10-06
BUGFIX: do collect the last row in the source table with pt-archiver

hjweddie:
1. When I archive source table to dest table using pt-archiver, I find that the last row is always missing, which means "last row is lost in the archive"
2. I think that we could keep the last row in the dest table/file when using pt-archiver


svetasmirnova:
This is not a bug but documented behavior. Check https://docs.percona.com/percona-toolkit/pt-archiver.html#cmdoption-pt-archiver--[no]safe-auto-increment and https://jira.percona.com/browse/PT-1349 for details.

PR #521, opened 2021-11-01
PT-2016 pt-table-sync and CRC32 columns

percona-csalguero:
pt-table-sync was not building queries correctly to loop using CRC32 hash function, on fields like text or blob.
The synchronization was working correctly on regular numeric PKs but in some places the `WHERE` clause was being written without the CRC32 function so the result was empty.
For example, if you have a table definition like this one:
```
CREATE TABLE test.test2 (
  col1 INT NOT NULL,
  col2 TEXT NOT NULL,
  col3 VARCHAR(5) NOT NULL,
  UNIQUE KEY (col1, col2(3))
);
```
The where clause should be `col1 = 'something' AND CRC32(col2) = 'something else'` because on text fields the checksum is being calculated using the `CRC32` function.
Also, enums and hex values where not properly quoted.


JiriCtvrtka (approved):
Just remove extra spaces

artemgavrilov (approved):
LGTM, but I'm not familiar with Perl.

PR #522, opened 2021-11-03
Updated makefile

percona-csalguero:
Updated makefile to:
- Use `./bin/` directory to store tools like `gofumpt` or linters
- Upgraded binaries
- Added `go mod tidy` to `make format`
- Updated ignored files list


src/go/Makefile
```diff
@@ -101,6 +101,12 @@ define TEST_ENV
 	TEST_MONGODB_FLAVOR=$(TEST_MONGODB_FLAVOR)
 endef
 
+init:                       ## Install linters.
+	go build -modfile=tools/go.mod -o bin/gofumports mvdan.cc/gofumpt/gofumports
+	go build -modfile=tools/go.mod -o bin/gofumpt mvdan.cc/gofumpt
+	go build -modfile=tools/go.mod -o bin/golangci-lint github.com/golangci/golangci-lint/cmd/golangci-lint
+	go build -modfile=tools/go.mod -o bin/reviewdog github.com/reviewdog/reviewdog/cmd/reviewdog
```
artemgavrilov, line 108: That can be implemented via `go:generate`
https://github.com/percona/pmm-managed/blob/222b46e62e9978a41b1f5eca39d20878ba6444e1/tools/tools.go#L37-L47

In that case you will have all tools settings in one package. But decision in up to you.

src/go/Makefile
```diff
@@ -136,15 +142,16 @@ darwin-amd64:				## Build Mongo tools for darwin-amd64 (MacOS)
 
 style:						## Check code style
 	@echo ">> checking code style"
-	@! gofmt -d $(shell find . -path ./vendor -prune -o -name '*.go' -print) | grep '^'
+	@! bin/gofumpt -d $(shell find . -path ./vendor -prune -o -name '*.go' -print) | grep '^'
```
artemgavrilov, line 145: 👍

src/go/Makefile
```diff
@@ -136,15 +142,16 @@ darwin-amd64:				## Build Mongo tools for darwin-amd64 (MacOS)
...
 test:						## Run tests
 	@echo ">> running tests"
 	@./runtests.sh
 
 format:                         ## Format source code.
-	gofumpt -w -s $(FILES)
-	gofumports -local github.com/percona/pmm-managed -l -w $(FILES)
+	go mod tidy
+	bin/gofumpt -w -s $(FILES)
+	bin/gofumports -local github.com/percona/pmm-managed -l -w $(FILES)
```
artemgavrilov, line 154: You can leave only `gofumports` because it's equal to `gofumt` + impors sorting

go.mod
```diff
@@ -28,11 +28,34 @@ require (
 
 require (
```
artemgavrilov, line 29: Why do you need a second `require` section that contains indirect dependencies? I'm almost sure that you can drop it completely.

src/go/tools/go.mod
```diff
@@ -0,0 +1,188 @@
...
+
+go 1.17
+
+require (
+	github.com/golangci/golangci-lint v1.43.0
+	github.com/reviewdog/reviewdog v0.13.0
+	golang.org/x/tools v0.1.7
+)
+
+require (
```
artemgavrilov, line 11: Same, IDK why you need this. All this deps pinned in `go.sum` file.

src/go/tools/go.mod
```diff
@@ -0,0 +1,188 @@
+module github.com/percona/mongodb_exporter/tools
+
+go 1.17
```
hors, line 3: can we use go 1.19?

percona-csalguero: yes

src/go/Makefile
artemgavrilov: Actually `gofumports` was deprecated. So you need `gofumpt + goimports`
https://github.com/percona/pmm/blob/9450d3d000fa7a35fbbdd12511688e4c5bff1165/Makefile.include#L163-L164

PR #533, opened 2022-01-19
PT-2037: option --system-only for pt-stalk

svetasmirnova:
Fix for PT-2037: Add option --skip-mysql or --system-only for pt-stalk

### DIFF
--- a/bin/pt-stalk
+++ b/bin/pt-stalk
@@ -894,7 +894,7 @@ collect_mysql_data_one() {
       tail -f "$mysql_error_log" >"$d/$p-log_error" &
       tail_error_log_pid=$!
 
-      $CMD_MYSQLADMIN $EXT_ARGV debug
+      $CMD_MYSQLADMIN $EXT_ARGV
    else
       log "Could not find the MySQL error log"
    fi 
@@ -934,8 +934,8 @@ collect_mysql_data_one() {
       have_lock_waits_table="yes"
    fi
 
-      $CMD_MYSQLADMIN $EXT_ARGV ext -i$OPT_SLEEP_COLLECT -c$cnt >>"$d/$p-mysqladmin" &
-      mysqladmin_pid=$!
+   $CMD_MYSQLADMIN $EXT_ARGV ext -i$OPT_SLEEP_COLLECT -c$cnt >>"$d/$p-mysqladmin" &
+   mysqladmin_pid=$!
 
    ps_instrumentation_enabled=$($CMD_MYSQL $EXT_ARGV -e 'SELECT ENABLED FROM performance_schema.setup_instruments WHERE NAME = "transaction";' \
                                       | sed "2q;d" | sed 'y/ABCDEFGHIJKLMNOPQRSTUVWXYZ/abcdefghijklmnopqrstuvwxyz/')
--- a/bin/pt-stalk
+++ b/bin/pt-stalk
@@ -796,6 +796,7 @@ CMD_DMESG="${CMD_DMESG:-"$(_which dmesg)"}"
 collect() {
    local d="$1"  # directory to save results in
    local p="$2"  # prefix for each result file
+   local cnt=$(($OPT_RUN_TIME / $OPT_SLEEP_COLLECT))
 
    if [ ! "$OPT_SYSTEM_ONLY" ]; then
       local mysqld_pid=""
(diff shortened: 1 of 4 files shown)


bin/pt-stalk
```diff
@@ -796,84 +797,88 @@ collect() {
    local d="$1"  # directory to save results in
    local p="$2"  # prefix for each result file
 
-   local mysqld_pid=""
-   if [ ! "$OPT_MYSQL_ONLY" ]; then
-      mysqld_pid=$(_pidof mysqld | awk '{print $1; exit;}')
-   fi
+   if [ ! "$OPT_SYSTEM_ONLY" ]; then
```
denisok, line 806: instead of `if` maybe it is good idea to create new function for system collect? and just call it in the beginning and exit if "$OPT_SYSTEM_ONLY" is set. 

And do we need new key `OPT_SYSTEM_ONLY` at all? if there is no mysqld process - just don't execute mysqld section, if OPT_MYSQL_ONLY is set don't execute the system function

svetasmirnova: I made pt-stalk data collection more modular in commit d6e89940..623fdaec. Please review.

Regarding the option: while the main purpose for this patch is to allow pt-stalk to run on machines where MySQL is not installed, we may need to run it on machines where MySQL is installed, but not the source of the problem. For example, working with a customer who has a small MySQL server and large MongoDB or PostgreSQL that we are troubleshooting.

bin/pt-stalk
```diff
@@ -128,7 +128,6 @@ OPT_ERRS=0        # How many command line option errors
 OPT_VERSION=""    # If --version was specified
 OPT_HELP=""       # If --help was specified
 OPT_ASK_PASS=""   # If --ask-pass was specified
-OPT_MYSQL_ONLY="" # If --mysql-only was specified
```
denisok, line 131: why did you remove this option ? you added another one so you should add it here. Probably it was missed from your previous commits.

bin/pt-stalk
```diff
@@ -459,7 +458,6 @@ _parse_command_line() {
          else
             spec=$(grep "^short form:-$opt\$" "$PT_TMPDIR"/po/* | cut -d ':' -f 1)
             if [ -z "$spec"  ]; then
-               option_error "Unknown option: $real_opt"
```
denisok, line 462: why is this removed?

bin/pt-stalk
```diff
@@ -796,9 +801,65 @@ collect() {
...
+         log "Removing empty file $file";
+         rm "$file"
+      fi
+   done
+}
+
+collect_mysql_data_one() {
    if [ ! "$OPT_MYSQL_ONLY" ]; then
-      mysqld_pid=$(_pidof mysqld | awk '{print $1; exit;}')
+      port=$($CMD_MYSQL $EXT_ARGV -ss -e 'SELECT @@port')
```
denisok, line 861: ok, is this some kind of improvement? Should we separate this to another ticket, so it could be more descriptive. Because adding it in the scope of this fix - it could be hard to figure out later why it was done.

bin/pt-stalk
```diff
@@ -820,27 +881,27 @@ collect() {
...
       mysql_error_log="$(ls -l /proc/$mysqld_pid/fd | awk '/ 2 ->/{print $NF}')"
    fi
 
-   local tail_error_log_pid=""
    if [ "$mysql_error_log" -a ! "$OPT_MYSQL_ONLY" ]; then
       log "The MySQL error log seems to be $mysql_error_log"
       tail -f "$mysql_error_log" >"$d/$p-log_error" &
       tail_error_log_pid=$!
 
-      $CMD_MYSQLADMIN $EXT_ARGV debug
```
denisok, line 836: why we delete this?

bin/pt-stalk
```diff
@@ -1171,11 +1219,11 @@ slave_status() {
       echo -e "\n$sql\n" >> $outfile
       $CMD_MYSQL $EXT_ARGV -e "$sql" >> $outfile
 
-      sql="SELECT * FROM replication_connection_status\G"
+      sql="SELECT * FROM performance_schema.replication_connection_status\G"
       echo -e "\n$sql\n" >> $outfile
       $CMD_MYSQL $EXT_ARGV -e "$sql" >> $outfile
 
-      sql="SELECT * FROM replication_applier_status JOIN replication_applier_status_by_coordinator USING(channel_name)\G"
+      sql="SELECT * FROM performance_schema.replication_applier_status JOIN performance_schema.replication_applier_status_by_coordinator USING(channel_name)\G"
```
denisok, line 1226: how did it work before ? or it will drop support for some old DB versions?

t/pt-stalk/pt-stalk.t
```diff
@@ -542,20 +542,68 @@ $retval = system("$trunk/bin/pt-stalk --no-stalk --run-time 2 --sleep 2 --dest $
...
+$output = $output - `ls -l $dest | wc -l`;
 
 is(
    $output,
    0,
    "Retention test 5: retention by count works as expected"
 );
 
+# ###########################################################################
+# Test if option --system-only works correctly
```
denisok, line 560: what if both options are given ? could you add test for that and check how your system behaves ?

denisok (commented):
Overall looks good, I have couple of questions.

bin/pt-stalk
svetasmirnova: This is the fix for https://jira.percona.com/browse/PT-1340 that was removed by the fix for PT-1398

bin/pt-stalk
svetasmirnova: It did not work. The log contained error messages that table replication_applier_status is not found. I believe this was, again, the result of mix-up of library and tool code.

bin/pt-stalk
svetasmirnova: Because this is part of the auto-copied code from lib/bash/parse_options.sh The tool works correctly without it

svetasmirnova: Function _eval_po from parse_options.sh evaluates such options automatically and assigns value of mysql-only to OPT_MYSQL_ONLY, value system-only to OPT_SYSTEM_ONLY, and so on

bin/pt-stalk
svetasmirnova: This is fix for https://jira.percona.com/browse/PT-1720 that was not added to pt-stalk, because modules were not updated properly

bin/pt-stalk
svetasmirnova: This is the fix for https://jira.percona.com/browse/PT-1398

t/pt-stalk/pt-stalk.t
svetasmirnova: Test case added. See commit 3767e57e151fe7e3f9631d5c736334c1b8fc431b

denisok (approved):

PR #538, opened 2022-04-03
fix Cannot find encoding "utf8mb4"

slowtech:
Here is a test case.

`
mysql> select @@version;
+------------+
| @@version  |
+------------+
| 5.7.36-log |
+------------+
1 row in set (0.00 sec)

mysql> show create table sbtest.t1\G
*************************** 1. row ***************************
       Table: t1
Create Table: CREATE TABLE `t1` (
  `id` int(11) NOT NULL,
  `c1` varchar(10) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4
1 row in set (0.00 sec)

mysql> show create table sbtest.t2\G
*************************** 1. row ***************************
       Table: t2
Create Table: CREATE TABLE `t2` (
  `id` int(11) NOT NULL,
  `c1` varchar(10) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4
1 row in set (0.00 sec)

mysql> select * from sbtest.t1;
+----+--------------+
| id | c1           |
+----+--------------+
|  1 | slowtech😊     |
+----+--------------+
1 row in set (0.00 sec)
`
then we archive data from t1 to t2.

./pt-archiver --source h=127.0.0.1,P=3307,u=root,p=123456,D=sbtest,t=t1 --dest h=127.0.0.1,P=3307,u=root,p=123456,D=sbtest,t=t2 --bulk-insert --limit 10 --where '1=1' --bulk-delete --charset utf8mb4
Cannot find encoding "utf8mb4" at ./pt-archiver line 6712.

The essential reason is that the file encoding cann't be ut8mb4.

### DIFF
--- a/bin/pt-archiver
+++ b/bin/pt-archiver
@@ -6709,6 +6709,9 @@ sub main {
       require File::Temp;
       $bulkins_file = File::Temp->new( SUFFIX => 'pt-archiver' )
          or die "Cannot open temp file: $OS_ERROR\n";
+      if ( $charset eq ":encoding(utf8mb4)" ) {
+          $charset = ":encoding(utf8)";
+      }
       binmode($bulkins_file, $charset)
          or die "Cannot set $charset as an encoding for the bulk-insert "
               . "file: $OS_ERROR";
@@ -6942,6 +6945,9 @@ sub main {
          if ( $o->get('bulk-insert') ) {
             $bulkins_file = File::Temp->new( SUFFIX => 'pt-archiver' )
                or die "Cannot open temp file: $OS_ERROR\n";
+            if ( $charset eq ":encoding(utf8mb4)" ) {
+               $charset = ":encoding(utf8)";
+            }
             binmode($bulkins_file, $charset)
                or die "Cannot set $charset as an encoding for the bulk-insert "
                     . "file: $OS_ERROR";


svetasmirnova (changes requested):

For us to accept it, you need to sign CLA. Please do it.

Please, also add a test case for your change.

percona-csalguero:
I am not sure if changing the variable value to `utf8` will work. 
It might work for this particular case but we need to add several test cases, with tables using `utf8` and `utf8mb4` and with MySQL 5.7 & MySQL 8.0+

svetasmirnova:
> I am not sure if changing the variable value to `utf8` will work. It might work for this particular case but we need to add several test cases, with tables using `utf8` and `utf8mb4` and with MySQL 5.7 & MySQL 8.0+

That character set in the tool's code is for the Perl module, not for MySQL. Perl understands such MySQL encodings as cp1251 but does not understand utf8mb4.

percona-csalguero:
> > I am not sure if changing the variable value to `utf8` will work. It might work for this particular case but we need to add several test cases, with tables using `utf8` and `utf8mb4` and with MySQL 5.7 & MySQL 8.0+
> 
> That character set in the tool's code is for the Perl module, not for MySQL. Perl understands such MySQL encodings as cp1251 but does not understand utf8mb4.

Exactly. That's why I am not sure about changing the value for the `charset` variable that might be used globally will work.
We should add several test cases to ensure we are not affecting the DB connection charset.

svetasmirnova:
@percona-csalguero , you are correct. It is safer to use the different variable name for the binmode code, because variable $charset is used in many other places.

@slowtech , please, adjust your PR

svetasmirnova:
Fixed the same issue differently, by https://github.com/percona/percona-toolkit/pull/705

PR #545, opened 2022-07-27
PT-2084 show slave status

guriandoro:
PT-2084 -- Re-adding show slave status outputs to pt-stalk.

Work done on PT-80 removed these outputs for versions 5.7+, which is not desirable (and the P_S outputs we get in them are not enough to get all data we have in slave status outputs).

Additionally, I'm adding a conditional for checking versions 8.1+ if they decide to remove the "slave" syntax:

https://dev.mysql.com/doc/refman/8.0/en/show-replica-status.html
>From MySQL 8.0.22, use [SHOW REPLICA STATUS](https://dev.mysql.com/doc/refman/8.0/en/show-replica-status.html) in place of [SHOW SLAVE STATUS](https://dev.mysql.com/doc/refman/8.0/en/show-slave-status.html), which is deprecated from that release.

Since ${mysql_version} is not using patch number, we can't compare on 8.0.22, but using 8.1 is a good workaround that will make this patch work in future releases too.

Let me know what you think.

### DIFF
--- a/t/pt-stalk/pt-stalk-replication.t
+++ b/t/pt-stalk/pt-stalk-replication.t
@@ -78,5 +78,5 @@ is(
 
 
 cleanup();
-#ok($sb->ok(), "Sandbox servers") or BAIL_OUT(__FILE__ . " broke the sandbox");
+ok($sb->ok(), "Sandbox servers") or BAIL_OUT(__FILE__ . " broke the sandbox");
 done_testing;
--- a/t/pt-stalk/pt-stalk-replication.t
+++ b/t/pt-stalk/pt-stalk-replication.t
@@ -1,0 +1,82 @@
+#!/usr/bin/env perl
+
+BEGIN {
+   die "The PERCONA_TOOLKIT_BRANCH environment variable is not set.\n"
+      unless $ENV{PERCONA_TOOLKIT_BRANCH} && -d $ENV{PERCONA_TOOLKIT_BRANCH};
+   unshift @INC, "$ENV{PERCONA_TOOLKIT_BRANCH}/lib";
+};
+
+use strict;
+use warnings FATAL => 'all';
+use threads;
+use English qw(-no_match_vars);
+use Test::More;
+use Time::HiRes qw(sleep);
+
+use PerconaTest;
+use DSNParser;
+use Sandbox;
+
+my $dp = new DSNParser(opts=>$dsn_opts);
+my $sb = new Sandbox(basedir => '/tmp', DSNParser => $dp);
+my $dbh = $sb->get_dbh_for('master');
+
+if ( !$dbh ) {
+   plan skip_all => 'Cannot connect to sandbox master';
+}
+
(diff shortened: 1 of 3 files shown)


svetasmirnova (changes requested):
SHOW REPLICA STATUS exists since 8.0.22, therefore check regarding 8.1 is wrong.
Also, there is no 8.1 version yet. Did you mean 8.0.1?
Also, please, add a test case for the change as described at https://github.com/percona/percona-toolkit/blob/3.x/CONTRIBUTE.md#setting-up-the-development-environment-and-testing

guriandoro:
Sveta,

Let me comment inline.

>SHOW REPLICA STATUS exists since 8.0.22, therefore check regarding 8.1 is wrong.

The ${mysql_version} variable pt-stalk uses to check versions only has major.minor values, so it's impossible for us to check if version is greater, equal or less than 8.0.22 in particular. We have to work with 5.7, 8.0, etc.

>Also, there is no 8.1 version yet. Did you mean 8.0.1?

I know there is no 8.1 yet. That part was being preemptive of future changes, in which since it's deprecated in 8.0.22, it will most likely be removed in 8.1+ in the future. This is to avoid the tool breaking in the future.

>Also, please, add a test case for the change 

Ok.

Thanks!

svetasmirnova:
OK, that makes sense regarding 8.1 Awaiting tests.

guriandoro:
Sveta,

New test case created as t/pt-stalk/pt-stalk-replication.t.
Please review it and let me know what you think.

```
shell> prove -vw --trap --timer t/pt-stalk/pt-stalk-replication.t
[22:05:50] t/pt-stalk/pt-stalk-replication.t ..
ok 1 - SHOW SLAVE STATUS outputs gathered.
ok 2 - Exit 0
1..2
ok     3954 ms ( 0.00 usr  0.00 sys +  0.75 cusr  1.39 csys =  2.14 CPU)
[22:05:54]
All tests successful.
Files=1, Tests=2,  4 wallclock secs ( 0.01 usr  0.01 sys +  0.75 cusr  1.39 csys =  2.16 CPU)
Result: PASS
```

t/pt-stalk/pt-stalk-replication.t
```diff
@@ -0,0 +1,82 @@
...
+   "Exit 0"
+);
+
+# #############################################################################
+# Done.
+# #############################################################################
+
+
+cleanup();
+#ok($sb->ok(), "Sandbox servers") or BAIL_OUT(__FILE__ . " broke the sandbox");
```
svetasmirnova:
```suggestion
ok($sb->ok(), "Sandbox servers") or BAIL_OUT(__FILE__ . " broke the sandbox");
```
We need to check if we left the sandbox in a consistent state.

svetasmirnova (approved):
We may need to fix test case in the future to check for REPLICA syntax as well.

percona-csalguero (approved):

PR #551, opened 2022-08-23
Add options to specify doc build target

denisok:
PR is needed to be able to build html or/and pdf

### DIFF
--- a/util/write-user-docs
+++ b/util/write-user-docs
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -x
+#set -x
 
 # Usage: write-user-docs [TOOLS]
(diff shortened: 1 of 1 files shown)


nastena1606:
Hi! Thanks for the PR. I've tested the build with --html flag but the script still builds both html and pdf docs.

denisok:
@nastena1606 you are using Mac that has old `getopt` unix version, which doesn't support long options. I changed it to short one, please try again. Also added `-t` for your request to have new make option in a future for new theme.

nastena1606 (approved):
LGTM, thank you!

percona-csalguero (approved):

PR #557, opened 2022-10-28
Update pt-kill

lirulei:
add User 、Host、DB column to log

### DIFF
--- a/bin/pt-kill
+++ b/bin/pt-kill
@@ -7409,9 +7409,9 @@ sub main {
                   }
                   printf "%s\n", Transformers::hash_to_json(\%jsonOutput);
                } else {
-                  printf "# %s %s %d (%s %d sec) %s\n",
-                     ts(time), $o->get('kill-query') ? 'KILL QUERY' : 'KILL',
-                     $query->{Id}, ($query->{Command} || 'NULL'), $query->{Time},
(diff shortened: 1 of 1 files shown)


cfsalguero:
Hi, thanks for the contribution.
This looks good but I wonder if the change brake any test.
Many tests in our test suite check the programs outputs.
Could you check the tests and if needed, update them?

Thanks

svetasmirnova (changes requested):
This patch has few issues.

**First**, pt-kill supports one of 5 actions:

1. **kill** that will close the connection
2. **kill-query** that will kill the query and do not close the connection
3. **print** that prints what the tool is going to do and can be used together with other actions
4. **execute-command** that executes custom command
5. **stop** that stops the daemon

For people who rely on logging of these actions they all are different. You are suggesting to do not distinguish between --kill-query and --kill, then ignore actions --execute and --stop. Also, you do not handle the case when --print is used together with other commands.

Current code prints if either KILL or KILL QUERY was used on line 7413 and prints the executed command if option --execute-command together with option --verbose was specified on line 7423. I believe this behavior is correct. I suggest either to handle all possible cases (or, at least, distinguish between KILL and KILL QUERY) or leave code on line 7413 as it is now.

**Second** issue is that adding addional output to the tool will break existing automated monitoring. I suggest to print such information only if option --verbose is used.

PR #558, opened 2022-11-16
PT 2105 collect individual logs

svetasmirnova:
This PR adds support for collecting individual log files from the PXC operator in addition to the combined log file in JSON format. This is required for easier manual troubleshooting of PXC-related issues.


cfsalguero (approved):

percona-csalguero (approved):

go.mod
```diff
@@ -1,8 +1,9 @@
 module github.com/percona/percona-toolkit
 
-go 1.17
+go 1.18
```
hors: I think we can use 1.19
```suggestion
go 1.19
```

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -60,6 +62,16 @@ func New(location, namespace, resource string) Dumper {
...
 				"perconaxtradbclusters")
+			filePaths = append(filePaths,
+				"var/lib/mysql/mysqld-error.log",
+				"var/lib/mysql/innobackup.backup.log",
+				"var/lib/mysql/innobackup.move.log",
+				"var/lib/mysql/innobackup.prepare.log",
+				"var/lib/mysql/grastate.dat",
+				"var/lib/mysql/gvwstate.dat",
+				"var/lib/mysql/mysqld.post.processing.log",
+				"var/lib/mysql/auto.cnf",
```
egegunes, line 73: any reason to not use absolute paths? currently we don't set `workingDir` in operators but if it's set, this will break.

egegunes (approved):
LGTM. I left a comment, it's up to you

src/go/pt-k8s-debug-collector/dumper/dumper.go
svetasmirnova: Because they are taken few times, for each of PXC nodes: cluster1-pxc-0, cluster1-pxc-1, and cluster1-pxc-2. So I have to dynamically create final absolute path.

egegunes: what worries me is we're using these paths as they're with `kubectl cp ...` and this can break with a `workingDir` field in PXC containers. but I don't think it's a problem right now, just room for improvement.

PR #568, opened 2022-12-27
PT-2138 - fix tests for pt table checksum

svetasmirnova:
Adjusted test cases in t/pt-table-checksum so they can run with the latest 5.7 and 8.0 versions.

- Updated t/pt-table-checksum/samples/default-results-8.0.txt and t/pt-table-checksum/samples/static-chunk-size-results-8.0.txt to support the latest 8.0 version
- Put fix for PT-136 into package RowChecksum;
- Added execution bit for pt-online-schema-change
- Fixed percona_test.load_data, so it tests if LOAD DATA LOCAL INFILE is enabled
- Fixed option --[no]create-replicate-table, broken by commit c9836d5962f27d5d24d1c28d2cf0066528d803b5
- Enabled t/pt-table-checksum/error_handling.t for MySQL 8.0
- Fixed test t/pt-table-checksum/fnv_64.t and it's samples file t/pt-table-checksum/samples/fnv64-sakila-city.txt to reflect new function name convention and changes after 62d84e5dba1669c0435225640c200f686be87d67
- Fixed t/pt-table-checksum/issue_1485195.t, so it checks only one table and isn't get broken when we add tables into the percona_test database
- Fixed typo in error output of bin/pt-table-checksum
- Skipped issue_47.t in 8.0 until https://jira.percona.com/browse/PT-1805 is fixed.
- Disabled pt-131.t for 8.0 because it does not have the QUERY_RESPONSE_TIME plugins
- Added SLOW_TESTS check to pt-1616.t
- Updated pt-226.t to include the fix for PT-1766
- For replication_filters.t: excluded false positive expression for tests 10 and 11 and added sys schema to the list of checked databases for 8.0
- Changed get_slaves in lib/[MasterSlave.pm](http://masterslave.pm/), so it returns slave's parent, required for wait_for_slaves in pt-table-checksum to work properly with chained slaves
- Modified pt-204.t  to support 8.0 and diffs in system tables due to timestamps
- Moved the fix for PT-1616 into the proper place: lib/[NibbleIterator.pm](http://nibbleiterator.pm/)
- pxc.t -added mysql.proxies_priv into ignore list because its timestamp is different on nodes
- pxc.t - removed FORK="pxc" from startup options for slave (non-cluster) nodes
- pxc.t - disabled wsrep replication with the help of the variable wsrep_on: sql_log_bin doesn't disable wsrep replication anymore. See https://jira.percona.com/browse/PXC-3464 for details
- Removed data.tar.gz from 5.7 sandbox configuration because it has an outdated definition for Performance Schema
- Disabled pxc.t for version 8.0 until PT-1699 is fixed
- start-sandbox - removed the first line (ALTER USER) from the init file because the following echo command rewrote it, and then repeated later.
- Adopted issue_1485195.t and basics.t for MyRocks-enabled setup
- replication_filters.t - added sys schema to the list of expected schemas for 5.7 and 8.0
- issue_1485195.t - added checks for the existence of mysql.plugin, func, and proxies_priv tables
- added samples/pt-131-wipe.sql that uninstalls the QRT plugin if this test earlier installed it
- adjusted return code in pt-204.t, because expected differences in mysql.proxies_priv


lib/PerconaTest.pm
```diff
@@ -635,6 +635,8 @@ sub no_diff {
    $test_diff = `diff $res_file $cmp_file 2>&1`;
    my $retval = $?;
 
+#REMOVEME
+diag(`diff $res_file $cmp_file`);
```
percona-csalguero:
```suggestion

```

lib/RowChecksum.pm
```diff
@@ -102,18 +102,30 @@ sub make_row_checksum {
       $sep    =~ s/'//g;
       $sep  ||= '#';
 
+      my @converted_cols;
+      for my $col(@{$cols->{select}}) {
+          my $colname = $col;
+          $colname =~ s/`//g;
+          my $type = $tbl_struct->{type_for}->{$colname} || '';
+          if ($type =~ m/^(CHAR|VARCHAR|BINARY|VARBINARY|BLOB|TEXT|ENUM|SET|JSON)$/i) {
```
percona-csalguero, line 110: Should we include `ENUM`s?
Enums are handled internally as an index number so most probably this is correct, but I am wondering if a conversion might affect how the column is being handled.
This would be a very particular case but, we could add a specific test.

denisok (approved):

percona-csalguero (approved):
LGTM with some minor comments.

Tusamarco (approved):
saw few comments removed in the code and additional spaces at the end of the lines ... but aside that seems all good to me

lib/RowChecksum.pm
svetasmirnova: ENUM is already here: if ($type =~ m/^(CHAR|VARCHAR|BINARY|VARBINARY|BLOB|TEXT|**ENUM**|SET|JSON)$/i) {

PR #576, opened 2023-01-24
PT-2164 version flag for pt-k8s-debug-collector

svetasmirnova:


egegunes (commented):
changes LGTM but `src/go/pt-k8s-debug-collector/main.go` needs `go fmt`

egegunes (approved):

denisok (approved):

PR #580, opened 2023-01-31
PT-2116 pt-osc --null-to-not-null description is wrong

ylacancellera:
Old documentation was incomplete: old rows would not use the new user defined default value for the altered column. Only new rows inserted after pt-osc would use the user defined default value. Existing rows would only be converted to a default value derived from the datatype, as it used to be the case with mysql 5.6

- [x] Documention updated
- [ ] Test suite update


svetasmirnova (approved):

alina-derkach-oaza (approved):
LGTM

PR #586, opened 2023-02-10
Create toolkit.yml

EvgeniyPatlan:
Added github action that will build go binaries on each commit. After that we will scan all binaries on CVEs. And if there are no CVEs binaries will be available for downloads

- [x] Documention updated
- [x] Test suite update

### DIFF
--- a/.github/workflows/toolkit.yml
+++ b/.github/workflows/toolkit.yml
@@ -23,12 +23,16 @@ jobs:
       run: cd src/go; make linux-amd64; cd ../../
       
     - name: Build the Docker image
-      run: echo "FROM oraclelinux:9-slim" > Dockerfile; echo "COPY bin/* /usr/bin/" >> Dockerfile; docker build . --file Dockerfile --tag percona-toolkit:$(date +%s)
-    - name: Install Trivy
-      run: wget https://github.com/aquasecurity/trivy/releases/download/v0.37.1/trivy_0.37.1_Linux-64bit.deb && sudo dpkg -i ./trivy_0.37.1_Linux-64bit.deb && rm -f trivy_0.37.1_Linux-64bit.deb
-    - name: Scan Image
-      run: trivy image --scanners vuln  --format table --exit-code  1 --vuln-type  os,library --severity  CRITICAL,HIGH percona-toolkit:$(docker images | grep percona-toolkit | head -n 1 | awk '{print $2}')
-     
+      run: echo "FROM oraclelinux:9-slim" > Dockerfile; echo "COPY bin/* /usr/bin/" >> Dockerfile; docker build . --file Dockerfile --tag percona-toolkit:${{ github.sha }}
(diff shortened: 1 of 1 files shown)


svetasmirnova (approved):

.github/workflows/toolkit.yml
```diff
@@ -0,0 +1,36 @@
...
+      run:  sudo apt-get update && sudo apt-get -y install make
+    - name: Build
+      run: cd src/go; make linux-amd64; cd ../../
+      
+    - name: Build the Docker image
+      run: echo "FROM oraclelinux:9-slim" > Dockerfile; echo "COPY bin/* /usr/bin/" >> Dockerfile; docker build . --file Dockerfile --tag percona-toolkit:$(date +%s)
+    - name: Install Trivy
+      run: wget https://github.com/aquasecurity/trivy/releases/download/v0.37.1/trivy_0.37.1_Linux-64bit.deb && sudo dpkg -i ./trivy_0.37.1_Linux-64bit.deb && rm -f trivy_0.37.1_Linux-64bit.deb
+    - name: Scan Image
+      run: trivy image --scanners vuln  --format table --exit-code  1 --vuln-type  os,library --severity  CRITICAL,HIGH percona-toolkit:$(docker images | grep percona-toolkit | head -n 1 | awk '{print $2}')
```
egegunes: it might be better use github action from aqua security. you can check PSMDB operator as an example: https://github.com/percona/percona-server-mongodb-operator/blob/main/.github/workflows/scan.yml#L16-L24

.github/workflows/toolkit.yml
```diff
@@ -0,0 +1,36 @@
...
+      uses: actions/setup-go@v3
+      with:
+        go-version: 1.19
+    - name: Install make
+      run:  sudo apt-get update && sudo apt-get -y install make
+    - name: Build
+      run: cd src/go; make linux-amd64; cd ../../
+      
+    - name: Build the Docker image
+      run: echo "FROM oraclelinux:9-slim" > Dockerfile; echo "COPY bin/* /usr/bin/" >> Dockerfile; docker build . --file Dockerfile --tag percona-toolkit:$(date +%s)
```
egegunes: we can use `${{ github.sha }}` rather than `$(date +%s)` as tag. example: https://github.com/percona/percona-server-mongodb-operator/blob/main/.github/workflows/scan.yml#L12

.github/workflows/toolkit.yml
EvgeniyPatlan: done

.github/workflows/toolkit.yml
EvgeniyPatlan: done

egegunes (approved):
LGTM but I didn't see the job runs and succeeds

EvgeniyPatlan:
> 

https://github.com/EvgeniyPatlan/percona-toolkit/actions/runs/4144875180

PR #604, opened 2023-03-09
Only rollback destination if it's defined

mikrobi:
This change addresses https://jira.percona.com/browse/PT-2064 which causes PT-archiver to crash when hitting an error. We use pt-archiver to delete rows that shouldn't be moved to another destination, so `--dest` is not provided.

In most places, we seem to check  if `$dst` is actually defined before using it, so we might want to do the same in the `do_with_retries` function? Here's one example where we check if `$dst` is defined:

https://github.com/percona/percona-toolkit/blob/79757865e36e2b0bf6ee214e10d77f0e2cc0982d/bin/pt-archiver#L6370

- [ ] Documention updated
- [ ] Test suite update


svetasmirnova (changes requested):
Please add a regression test. You will find examples in the t directory and user manual in the contribution guide.

svetasmirnova (approved):

EvgeniyPatlan (approved):

PR #615, opened 2023-04-15
PT-2169 - pt-k8s-debug-collector integration of pg_gather requires croping first line of the output file

svetasmirnova:
Modified pt-k8s-debug-collector so it redirects only STDOUT to summary.txt STDERR is stored in the logs and recorded in summary.txt only if summary fails with the error Modified Makefile, so it does not include closing bracket into the version string Created test case for the fix

- [ ] Documentation updated
- [x] Test suite update

### DIFF
--- a/src/go/pt-k8s-debug-collector/dumper/dumper.go
+++ b/src/go/pt-k8s-debug-collector/dumper/dumper.go
@@ -6,7 +6,6 @@ import (
 	"compress/gzip"
 	"encoding/base64"
 	"encoding/json"
-	"fmt"
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


percona-csalguero (approved):

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -463,10 +462,10 @@ func (d *Dumper) getPodSummary(resource, podName, crName string, namespace strin
 	cmd.Stderr = &errb
 	err := cmd.Run()
 	if err != nil {
-		return nil, errors.Errorf("error: %v, stderr: %s, stdout: %s", err, errb.String(), outb.String())
+		return nil, errors.Errorf("error: %v\nstderr: %sstdout: %s", err, errb.String(), outb.String())
+	} else {
+		return outb.Bytes(), nil
```
egegunes: this else is not needed, you can just `return outb.Bytes(), nil`

egegunes (changes requested):

egegunes (approved):

PR #652, opened 2023-07-28
PT-2236 - pt-secure-collect, pt-pg-summary do not follow PT standard

svetasmirnova:
…for option --version

- Redirected UsageWriter to os.Stdout for kingpin, so --version output goes to STDOUT, not to STDERR
- Adjusted text, printed by the --version flag
- Added test cases to check how --version flag works
- Adjusted test cases, so they use TOOLNAME constant

- [x] Documention updated
- [x] Test suite update

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -29,6 +29,7 @@ require (
 )
 
 require (
+	github.com/alecthomas/kingpin/v2 v2.3.2 // indirect
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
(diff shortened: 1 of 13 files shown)


src/go/pt-pg-summary/main.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/percona/percona-toolkit/src/go/pt-pg-summary/templates"
 )
 
+const (
+	TOOLNAME = "pt-pg-summary"
```
artemgavrilov: Go convention doesn't use all caps names for constants: https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps

It should be `Toolname` if it's public const, or most likely just `toolname`

go.mod
```diff
@@ -29,6 +29,7 @@ require (
 )
 
 require (
+	github.com/alecthomas/kingpin/v2 v2.3.2 // indirect
```
artemgavrilov: Please run command `go mod tidy`, it looks like this dependency id redundant.

svetasmirnova: Done

src/go/pt-pg-summary/main.go
svetasmirnova: Done

artemgavrilov (approved):

egegunes (approved):

PR #654, opened 2023-08-03
PT-2248 - pt-k8s-debug-collector does not run pg_gather with K8SPG 2

svetasmirnova:
- Added check for K8SPG 2, so can run pg_gather for it
- Added new allowed value for option --resource: -- pgv2 for K8SPG 2
-- auto to auto-detect custom resource
- Option --resource has now default value "auto"
- Updated documentation
- Added test cases for new options

- [x] Documention updated
- [x] Test suite update


src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -415,7 +444,38 @@ func (d *Dumper) getPodSummary(resource, podName, crName string, namespace strin
...
+			port = "5432"
+		}
+		cr, err := d.getCR("perconapgclusters", namespace)
+		if err != nil {
+			return nil, errors.Wrap(err, "get cr")
+		}
+		if cr.Spec.SecretName != "" {
+			user, err = d.getDataFromSecret(cr.Spec.SecretName, "username", namespace)
+		} else {
+			user, err = d.getDataFromSecret(crName+"-pguser-cluster1", "user", namespace)
```
egegunes: the secret name depends on the cluster name

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -415,7 +444,38 @@ func (d *Dumper) getPodSummary(resource, podName, crName string, namespace strin
...
+		}
+		if err != nil {
+			return nil, errors.Wrap(err, "get user from PostgreSQL users secret")
+		}
+		if cr.Spec.SecretName != "" {
+			pass, err = d.getDataFromSecret(cr.Spec.SecretName, "password", namespace)
+		} else {
+			pass, err = d.getDataFromSecret(crName+"-pguser-cluster1", "password", namespace)
```
egegunes: the secret name depends on the cluster name

egegunes (changes requested):

src/go/pt-k8s-debug-collector/dumper/dumper.go
svetasmirnova: So it should be something like
```suggestion
			user, err = d.getDataFromSecret(crName+"-pguser-"+crName, "user", namespace)
```
?

src/go/pt-k8s-debug-collector/dumper/dumper.go
svetasmirnova: So this should be something like
```suggestion
			pass, err = d.getDataFromSecret(crName+"-pguser-"+crName, "password", namespace)
```
?

egegunes: yes, but I want to mention that it's not guaranteed to have this user in cluster. if the end user configured `spec.users`, operator doesn't create this user. but we plan to change this logic and have a default user. @nmarukovich am i right?

svetasmirnova: Could you also please confirm if currently this user name is created as:

`<CLUSTER NAME>-<pguser of anything, specified in spec.users>-<CLUSTER NAME>`

So for default configuration this is cluster1-pguser-cluster1 and could be, for example mysupercluster-mysuperuser-mysupercluster? Or other variations are also possible?

egegunes: by default it's `<CLUSTER NAME>-pguser-<CLUSTER NAME>`
if an user is specified in `spec.users` it'll be `<CLUSTER NAME>-pguser-<USERNAME>`
and please note that the secret name is customizable with `spec.users[].secretName`

svetasmirnova: Thanks, implemented! Please review the updated patch.

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -53,56 +63,72 @@ func New(location, namespace, resource string, kubeconfig string, forwardport st
...
+			log.Panicf("Cannot get API resources and option --resource=auto specified:\n%s", err)
+		}
+		matches := resourcesRe.FindAllStringSubmatch(string(result), -1)
+		if len(matches) == 0 {
+			resource = "none"
+			break
 		}
+		for _, match := range matches {
+			resources = append(resources, match[1])
+			resource = match[2]
```
ylacancellera, line 80: Not sure if I'm getting it right, but:
does it mean "auto" will try to match a single resource ?

If it matches both pxc+pg, resource will (maybe) trigger on pxc, then pg, so resource will be overwritten with "pg"

Later on, it means it won't be adding the proper resources down the line ?

What I'm describing would only be useful if customer create different db in the same namespace, but here's my counter proposal:
creating a map such as
```
type resourceData struct {
  k8sResourcesToCheck []string
  fileToCheck []string
}
var resourcesMap map[string]resourceData
```
Later on, use that as
```
for _, match := range matches {
  resources = append(resources, resourcesMap[match[1]].k8sResourcesToCheck...)
  filePaths = append(resources, resourcesMap[match[1]].filesToCheck...)
}
```
And for specific resource, the whole switch case could be replaced with
```
resources = append(resources, resourcesMap[resource].k8sResourcesToCheck...)
filePaths = append(resources, resourcesMap[resource].filesToCheck...)
```

(with some handling to avoid nil slice being appended)
That way, adding a resource and/or a file to check would just need completing this map

svetasmirnova: Hm... I didn't think someone would run two resources; if they did, there is always an option to specify the correct resource explicitly. Do you think it would be the case when the user has two resources, and we still need data from both (or more) of them?

ylacancellera: I thought about a scenario where some customers had a "project" namespace, like "reporting", with maybe multiple technologies under same namespace.

If there's a pxc + a pg cluster, I think one could be missed with "resource=auto" currently. If we miss the one we wanted, it's a bit disappointing
We can always explicitly set -resource sure, but IMO it's better when the tool accommodate it by catching everything that could be relevant right away, as I'm sure some will miss/forget to add the -resource parameter

The map solution could enable to catch everything, and would reduce the amount of if/else/switch code.
It would replace it with something like
```
resourcesMap := map[string]resourceData{
    "pxc": resourceData{ 
        filesToCheck: []string{"var/lib/mysql/mysqld-error.log", <and so on>}, 
        k8sResourcesToCheck : []string{"perconaxtradbclusterbackups", "perconaxtradbclusterrestores", "perconaxtradbclusters"},
    },
    "pg": ....
```

I'm thinking of a map here because it does not look so easy to add a loop with the current code to add multiple resources at once
I think it can avoid potential back and forth to investigate, and would enable to avoid touching code when modifying/adding resources later on

ylacancellera: Also: I can draft that code of course if needed
EDIT: actually now that I check it, my proposition should be in another PR anyway, not in this one. So, I'll just make a PR myself and see if it's useful

egegunes (approved):

artemgavrilov (approved):

src/go/pt-k8s-debug-collector/dumper/dumper.go
svetasmirnova: Thank you, @ylacancellera ! Let's make it another Jira and PR.

PR #676, opened 2023-09-11
PT-2168 pt-osc shouldnt fail while unable to monitor a replica node

svetasmirnova:
- [ ] Documentation updated
- [ ] Test suite update

Our user manual says at https://docs.percona.com/percona-toolkit/pt-online-schema-change.html:

> The tool waits forever for replicas to stop lagging. If any replica is stopped, the tool waits forever until the replica is started. The data copy continues when all replicas are running and not lagging too much.

So, actually, PT-2168 is not a feature request but a bug. However, at least since the year 2019 and #420, pt-osc fails on the stopped replica. Therefore, I implemented an option similar to one that pt-tc has: --fail-on-stopped-replication but made it disabled by default, so the behavior of last years is not broken.


EvgeniyPatlan (approved):

saikumar-vs (approved):
Nice patch. LGTM.

PR #686, opened 2023-09-24
Pt 2247

Tusamarco:
- [ x] The contributed code is licensed under GPL v2.0
- [ x] Contributor Licence Agreement (CLA) is signed
- [ NA] util/update-modules has been ran
- [ x] Documentation updated
- [NA ] Test suite update

@svetasmirnova please check and review also https://jira.percona.com/browse/PT-2247


svetasmirnova (changes requested):
PLease add test case. You can find how in the contribution guide and examples in t/pt-show-grants/

bin/pt-show-grants
```diff
@@ -2039,25 +2039,52 @@ sub main {
 
       # If MySQL 5.7.6+ then we need to use SHOW CREATE USER
       my @create_user;
-      if (( VersionCompare::cmp($version, '5.7.6') >= 0 ) &&
+      if (( VersionCompare::cmp($version, '5.7.6') >= 0 ) ||
```
svetasmirnova: With "||" this expression is always true.

svetasmirnova (changes requested):

bin/pt-show-grants
```diff
@@ -2433,6 +2467,16 @@ type: array
 
 Only show grants for this comma-separated list of users.
 
+=item --convert-MariaDB
+
+When set it convert some of the proprietary MariaDB syntax into valid MySQL form
```
svetasmirnova: Can't we do this automatically if we connect to MariaDB? Or will it break for valid MySQL syntax?

bin/pt-show-grants
```diff
@@ -2433,6 +2467,16 @@ type: array
...
+
+=item --print_identified_with_as_hex
```
svetasmirnova: According to the user manual at https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_print_identified_with_as_hex:

> Hash values that do not contain unprintable characters still display as regular string literals, even with this variable enabled. 

Do we really need this option or checking for correct MySQL version is enough?

bin/pt-show-grants
svetasmirnova: Looks like not, useful option.

svetasmirnova:
@Tusamarco I changed your PR: simplified a few things, removed the option --print_identified_with_hex (made it default and only behavior), and added a test case. Please review and see if you agree with the changes.

Tusamarco (commented):
LGTM

svetasmirnova (approved):

EvgeniyPatlan (approved):

PR #689, opened 2023-09-27
Bugfix - caching_sha2_passwords non-printable characters

fmbiete:
Authentication string for caching_sha2_passwords can contain non-printable characters, to generate a safe string is recommended to use always the binary representation of it.

More information:
https://lefred.be/content/cut-paste-a-user-creation-statement-with-mysql-8/

- [x] Documentation updated
- [x] Test suite update

### DIFF
--- a/bin/pt-show-grants
+++ b/bin/pt-show-grants
@@ -2037,6 +2037,12 @@ sub main {
          PTDEBUG && _d('Checking user', $user_host);
       }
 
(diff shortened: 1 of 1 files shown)


svetasmirnova (changes requested):
Please add a test case. You can find small how-to in CONTRIBUTING.md and examples in t/pt-show-grants/

svetasmirnova:
We will implement it as part of https://jira.percona.com/browse/PT-2247

PR #695, opened 2023-10-04
Bugfix pt-archiver miss the last row in source table

hjweddie:
I find pt-archiver is missing the last row

- [ ] Documentation updated
- [ ] Test suite update


svetasmirnova (changes requested):

Please add a test case. You will find a short how-to in the CONTRIBUTING.md and examples in t/pt-archiver/.

hjweddie:
I find that I makde a mistake understanding --no-delete and --[no]safe-auto-increment options. Now I think we do not need this pr at all

PR #696, opened 2023-10-04
PT-2109 - pt-mysql-summary throws an error when using sql_mode="ANSI_QUOTES"

svetasmirnova:
…QUOTES"

- Fixed issues still visible with SQL mode ANSI_QUOTES
- Added test cases for this mode

- [ ] Documentation updated
- [ ] Test suite update

### DIFF
--- a/t/pt-mysql-summary/pt-2109.t
+++ b/t/pt-mysql-summary/pt-2109.t
@@ -1,0 +1,80 @@
+#!/usr/bin/env perl
+
+BEGIN {
+   die "The PERCONA_TOOLKIT_BRANCH environment variable is not set.\n"
+      unless $ENV{PERCONA_TOOLKIT_BRANCH} && -d $ENV{PERCONA_TOOLKIT_BRANCH};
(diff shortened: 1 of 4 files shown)


t/pt-mysql-summary/pt-2109.t
```diff
@@ -0,0 +1,80 @@
...
+    while (my ($key, $values) = each %$rows) {
+        if ($key =~ m/^keyring_/) {
+            $has_keyring_plugin=1;
+            last;
+        }
+    }
+}
+
+if (!$has_keyring_plugin) {
+    plan skip_all => 'Keyring plugins are not enabled.';
```
saikumar-vs: Do we really need to skip this test if "Keyring" plugin is not enabled?

EvgeniyPatlan (approved):

t/pt-mysql-summary/pt-2109.t
svetasmirnova: Good point! Removed.

saikumar-vs (approved):

PR #697, opened 2023-10-06
PT-2254 pt-k8s-debug-collector: -resource=auto should get every possi…

ylacancellera:
…ble resources at once

Instead of relying on a "passthrough" mechanism on switch case, declaring a mapping between api and linked resources is easier and enable to iterate easily. It also ease contributions, as adding files/resources would not touch code anymore

- [ ] Documentation updated
- [ ] Test suite update

### DIFF
--- a/src/go/pt-k8s-debug-collector/dumper/dumper.go
+++ b/src/go/pt-k8s-debug-collector/dumper/dumper.go
@@ -10,7 +10,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
 	"strings"
 	"time"
 
@@ -32,8 +31,6 @@ type Dumper struct {
 	forwardport string
 }
 
-var resourcesRe = regexp.MustCompile(`(\w+)\.(\w+).percona\.com`)
-
 // New return new Dumper object
 func New(location, namespace, resource string, kubeconfig string, forwardport string) Dumper {
 	d := Dumper{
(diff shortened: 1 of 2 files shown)


ylacancellera:
Manual test suceeds, but actual tests fails:
```
--- FAIL: TestResourceOption (920.40s)
    main_test.go:212: test auto pxc, output is not as expected
        Output: 0
        
        Wanted: 3
    main_test.go:212: test auto ps, output is not as expected
        Output: 0
        
        Wanted: 3
    main_test.go:212: test auto psmdb, output is not as expected
        Output: 0
        
        Wanted: 3
    main_test.go:212: test auto pg, output is not as expected
        Output: 0
        
        Wanted: 3
    main_test.go:212: test auto pgv2, output is not as expected
        Output: 0
        
        Wanted: 3

```
Not sure about my test setup: I have k8s using anydbver, with every operators installed on separate namespaces.
For every auto builds, I actually have every namespaces everytime, it seems like a kubeconfig issue on my end

I don't get the "output" test, here's what I have when testing manually
```
yoann-lc@yoannlc-lemur:~/repo/percona-toolkit/src/go/pt-k8s-debug-collector$ ../../../bin/pt-k8s-debug-collector --kubeconfig kubeconfig_pg --forwardport 12456 --resource auto
2023/10/06 11:45:04 Start collecting cluster data
2023/10/06 11:46:35 Done
```
But, cluster.dump.tar.gz will have too many data. kubectl api-resources and kubectl api-versions will send every Percona resources, cluster-wide, so cluster.dump.tar.gz is currently having a yaml for each possible pxc/ps/pg/pgv2/psmdb resources for each namespaces.
So that currently gives me 378 files on the test above (without --namespace, which is how go tests are made)

PR #706, opened 2023-11-07
Follow PR-429 unstop option with testcase

svetasmirnova:
- [ ] Documentation updated
- [ ] Test suite update

### DIFF
--- a/bin/pt-archiver
+++ b/bin/pt-archiver
@@ -6244,6 +6244,8 @@ sub main {
    if ( $o->get('unstop') ) {
       unlink $sentinel 
          or die "Cannot unlink sentinel file: $OS_ERROR\n";
+      print STDOUT "Successfully removed file $sentinel\n"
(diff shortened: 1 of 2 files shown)


bin/pt-archiver
```diff
@@ -6240,6 +6240,14 @@ sub main {
       return 0;
    }
 
+   # Remove sentinel file.
+   if ( $o->get('unstop') ) {
+      unlink $sentinel 
+         or die "Cannot unlink sentinel file: $OS_ERROR\n";
```
EvgeniyPatlan: I would suggest 
"Cannot unlink $sentinel file: $OS_ERROR\n";

EvgeniyPatlan (changes requested):

EvgeniyPatlan (approved):

saikumar-vs (approved):

PR #716, opened 2023-11-17
Make OrderedBy not exported

szepeviktor:
- [ ] Documentation updated
- [ ] Test suite update

Make OrderedBy's return type exported as OrderedBy is exported.

### DIFF
--- a/src/go/pt-mongodb-query-digest/main.go
+++ b/src/go/pt-mongodb-query-digest/main.go
@@ -407,32 +407,32 @@ func getTotalsTemplate() string {
 
 type lessFunc func(p1, p2 *stats.QueryStats) bool
 
-type multiSorter struct {
+type MultiSorter struct {
 	queries []stats.QueryStats
 	less    []lessFunc
 }
 
 // Sort sorts the argument slice according to the less functions passed to OrderedBy.
-func (ms *multiSorter) Sort(queries []stats.QueryStats) {
+func (ms *MultiSorter) Sort(queries []stats.QueryStats) {
 	ms.queries = queries
 	sort.Sort(ms)
 }
 
 // OrderedBy returns a Sorter that sorts using the less functions, in order.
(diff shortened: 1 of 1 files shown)


svetasmirnova:
What is the reason of this change? main.go is not library code. Where are you going to use this structure?

szepeviktor:
I've just discovered
```go
func OrderedBy(less ...lessFunc) *multiSorter {
```
that an exported function has a not exported return type.

src/go/pt-mongodb-query-digest/main.go
```diff
@@ -407,32 +407,32 @@ func getTotalsTemplate() string {
...
+func (ms *MultiSorter) Sort(queries []stats.QueryStats) {
 	ms.queries = queries
 	sort.Sort(ms)
 }
 
 // OrderedBy returns a Sorter that sorts using the less functions, in order.
 // Call its Sort method to sort the data.
-func OrderedBy(less ...lessFunc) *multiSorter {
-	return &multiSorter{
+func OrderedBy(less ...lessFunc) *MultiSorter {
```
svetasmirnova: Sort, Len, Swap, Less are methods of the sort interface: https://pkg.go.dev/sort#Interface I suggest to leave name multiSorter as is and rename OrderedBy instead:

```suggestion
func orderedBy(less ...lessFunc) *multiSorter {
```

svetasmirnova (changes requested):

svetasmirnova:
Also found this sample: https://devbits.app/x/761/using-sort-sortmultikeys-in-go

src/go/pt-mongodb-query-digest/main.go
szepeviktor: Done 🍏

artemgavrilov (approved):

svetasmirnova (approved):

PR #717, opened 2023-11-17
PT-2281 - provide container name for copying files in the dump

svetasmirnova:
- Changed getIndividualFiles function and Dumper data structure, so we can specify container name for PXC and other operators which store logs in the separate container.

- [ ] Documentation updated
- [ ] Test suite update


src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -372,7 +374,7 @@ type crSecrets struct {
 
 // TODO: check if resource parameter is really needed
 func (d *Dumper) getIndividualFiles(resource, namespace string, podName, path, location string, tw *tar.Writer) error {
-	args := []string{"-n", namespace, "cp", podName + ":" + path, "/dev/stdout"}
+	args := []string{"-n", namespace, "-c", d.fileContainer, "cp", podName + ":" + path, "/dev/stdout"}
```
cshiv, line 379: As discussed, `d.fileContainer` is used for pxc only which means if this function is called in future for PSMDB/PG, this is effectively calling `kubectl -c cp` which means an empty container name and the command will error out.  It would be good to consider this case and fix it, so that getIndividualFiles can work independent of any constraints

artemgavrilov (approved):

src/go/pt-k8s-debug-collector/dumper/dumper.go
svetasmirnova: @cshiv, how do you suggest to handle it? Return error or use default container?

cshiv: @svetasmirnova Using the container name is an assured way and I believe we should continue using container name for getting the file. I would prefer returning an error.

svetasmirnova: Implemented, please review.

egegunes (approved):

cshiv (approved):
LGTM

PR #734, opened 2023-12-13
PT-2100 - fixed Wide character error

samitani:
- [] util/update-modules has been ran
- [] Documentation updated
- [] Test suite update

- fixed https://jira.percona.com/browse/PT-2100
- `use utf8` was introduced in  https://github.com/percona/percona-toolkit/commit/2233224ff67565b452b58fe3386768626b1a46ab. we should remove wide characters from output to avoid `Wide character error` under `use utf8`. I've replaced wide character `’` to single character `'`.


svetasmirnova:
I believe this is fixed by this commit: https://github.com/percona/percona-toolkit/commit/2233224ff67565b452b58fe3386768626b1a46ab that adds `use utf8;` into the tool's code. I cannot repeat the issue with the current version 3.5.7. So this PR only fixes how the word "tool's" is written. I will ask second opinion from our English experts before accepting it.

svetasmirnova (approved):

bin/pt-online-schema-change
```diff
@@ -8706,7 +8706,7 @@ sub main {
 
       # Issue a strong warning if alter-foreign-keys-method = none
       if ( $alter_fk_method eq 'none' && !$o->get('force') ) {
-         print STDERR "WARNING! Using alter-foreign-keys-method = \"none\". This will typically cause foreign key violations!\nThis method of handling foreign key constraints is only provided so that the database administrator can disable the tool’s built-in functionality if desired.\n\nContinue anyway? (y/N)";
+         print STDERR "WARNING! Using alter-foreign-keys-method = \"none\". This will typically cause foreign key violations!\nThis method of handling foreign key constraints is only provided so that the database administrator can disable the tool's built-in functionality if desired.\n\nContinue anyway? (y/N)";
```
patrickbirch: I suggest that this be rewritten to remove the apostrophe. 
Change to:

database adminstrator can disable the built-in functionality if

svetasmirnova:
```suggestion
         print STDERR "WARNING! Using alter-foreign-keys-method = \"none\". This will typically cause foreign key violations!\nThis method of handling foreign key constraints is only provided so that the database administrator can disable the built-in functionality if desired.\n\nContinue anyway? (y/N)";
```

patrickbirch (approved):
good with the fixed sentence

PR #748, opened 2023-12-26
Configure pt-archiver to archive the last row of a table

JensVD:
- Now pt-archiver can archive all but one row from a table. With this small fix it is able to archive all data from a table

- [x] Documentation updated
- [ ] Test suite update

### DIFF
--- a/bin/pt-archiver
+++ b/bin/pt-archiver
@@ -6260,7 +6260,7 @@ sub main {
 
    # Remove sentinel file.
    if ( $o->get('unstop') ) {
-      unlink $sentinel 
+      unlink $sentinel
          or die "Cannot unlink sentinel file ${sentinel}: $OS_ERROR\n";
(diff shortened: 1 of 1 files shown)


svetasmirnova:
This is expected behavior. Please see description of the option [--[no]safe-auto-increment](https://docs.percona.com/percona-toolkit/pt-archiver.html#cmdoption-pt-archiver-no-safe-auto-increment):

> default: yes
> 
>     Do not archive row with max AUTO_INCREMENT.
> 
>     Adds an extra WHERE clause to prevent pt-archiver from removing the newest row when ascending a single-column AUTO_INCREMENT key. This guards against re-using AUTO_INCREMENT values if the server restarts, and is enabled by default.
> 
>     The extra WHERE clause contains the maximum value of the auto-increment column as of the beginning of the archive or purge job. If new rows are inserted while pt-archiver is running, it will not see them.

PR #750, opened 2024-01-08
PT-2298 pt-galera-log-explainer superfluous lines on conflicts

ylacancellera:
- [x] Test suite update

### DIFF
--- a/src/go/pt-galera-log-explainer/regex/applicative.go
+++ b/src/go/pt-galera-log-explainer/regex/applicative.go
@@ -93,14 +93,15 @@ var ApplicativeMap = types.RegexMap{
 			return logCtx, func(logCtx types.LogCtx) string {
 
 				for _, localname := range logCtx.OwnNames {
-					vote, ok := latestConflict.VotePerNode[localname]
-					// don't print other nodes vote, their should be in their own column
-					// else it just takes too much place without adding much usability
-					if node != localname || !ok {
+					if node != localname {
+						// don't print other nodes vote, their should be in their own column
+						// else it just takes too much place without adding much usability
 						continue
 					}
 
-					return voteResponse(vote, *latestConflict)
+					if vote, ok := latestConflict.VotePerNode[localname]; ok {
+						return voteResponse(vote, *latestConflict)
+					}
 				}
 
 				return ""
--- a/src/go/pt-galera-log-explainer/regex/applicative.go
+++ b/src/go/pt-galera-log-explainer/regex/applicative.go
@@ -92,9 +92,11 @@ var ApplicativeMap = types.RegexMap{
 
 			return logCtx, func(logCtx types.LogCtx) string {
 
-				for _, name := range logCtx.OwnNames {
-					vote, ok := latestConflict.VotePerNode[name]
-					if !ok {
+				for _, localname := range logCtx.OwnNames {
+					vote, ok := latestConflict.VotePerNode[localname]
+					// don't print other nodes vote, their should be in their own column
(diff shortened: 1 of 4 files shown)


svetasmirnova (commented):
What is the difference this change makes? I see the same output for the test log file with and without this change.

ylacancellera:
Instead of
```
2023-10-21T04:01:01.700706Z   inconsistency vote started(seqno:102573168)   
                              consistency vote(seqno:102573168): lost    
```
It would have shown
```
2023-10-21T04:01:01.700706Z   inconsistency vote started(seqno:102573168)
2023-10-21T04:01:01.700706Z   inconsistency vote started(seqno:102573168)           
2023-10-21T04:01:01.701416Z   consistency vote(seqno:102573168): voted same error   
2023-10-21T04:01:01.701436Z   consistency vote(seqno:102573168): voted same error   
                              consistency vote(seqno:102573168): lost           
```
Both "voted same error" lines are a bug here. The log line is the other nodes response to the conflict, but the loop was always matching the local node. So it's a "double" bug, those lines should not have been here and the output is wrong anyway

It's not obvious with a single log here, but each logs will only vote its own response

svetasmirnova (approved):

svetasmirnova:
Thank you for the feedback, approved!

src/go/pt-galera-log-explainer/types/timeline.go
```diff
@@ -139,7 +139,7 @@ func getlasttime(l LocalTimeline) time.Time {
 func CutTimelineAt(t LocalTimeline, at time.Time) LocalTimeline {
 	var i int
 	for i = 0; i < len(t); i++ {
-		if t[i].Date.Time.After(at) {
+		if t[i].Date != nil && t[i].Date.Time.After(at) {
```
artemgavrilov, line 142: Are you sure that this is correct behavior? Cut timeline once we met nil value?

src/go/pt-galera-log-explainer/regex/applicative.go
```diff
@@ -92,9 +92,11 @@ var ApplicativeMap = types.RegexMap{
 
 			return logCtx, func(logCtx types.LogCtx) string {
 
-				for _, name := range logCtx.OwnNames {
-					vote, ok := latestConflict.VotePerNode[name]
-					if !ok {
+				for _, localname := range logCtx.OwnNames {
```
artemgavrilov, line 95: Minor: I would suggest some refactoring to this chunk
```
				for _, localname := range logCtx.OwnNames {
					if node != localname {
						// don't print other nodes vote, their should be in their own column
						// else it just takes too much place without adding much usability
						continue
					}

					if vote, ok := latestConflict.VotePerNode[localname]; ok {
						return voteResponse(vote, *latestConflict)
					}
				}
```

artemgavrilov (approved):

src/go/pt-galera-log-explainer/regex/applicative.go
ylacancellera: Agree, way better your way

src/go/pt-galera-log-explainer/types/timeline.go
ylacancellera: This is the correct behavior: it's searching the first element which happened just after the date in parameter
It's just that few logs don't have a "date", so it's nil. It's only a silly mistake I'm fixing,` .Date` is a` *types.Date`

(they don't have a date because galera is making multiline logs, and multiline regexes are not worth it)

PR #765, opened 2024-02-09
pt-galera-log-explainer: improvments from feedbacks

ylacancellera:
Don't get scared by +6482 lines, most of it is from adding a regression test (locally generated)
To me commits are short enough, let me know if that's too much. 

summary of various changes:

- "grep -a": because some logs are sent in a binary format, even though it's actually readable.
- --skip-merge: useful for cases with identical wsrep node names (got one in actual customer case)
- small refactor of "extractor" sub-logger. It was not using the same params as the general one, it was giving 2 formats of log when using -vv

operators log improvements: 
- auto detection because many people missed --pxc-operator
- setting pod name as identifier because it make more sense than using filenames
- enabling a merge mechanism because many customer cases ends up having many pt-k8s-debug-collector with completely different log files when pvc is destroyed. (merge uses podname+deployment+namespace)

**Important**:
- logs have been added in b15cdf2, but they were generated using a local k3d setup.
- make sure to review 3a5bde7, this is a debatable change. (https://github.com/alecthomas/kong/issues/408)

If this format of PR is not ideal let me know. 

- [x] Documentation updated
- [x] Test suite update

### DIFF
--- a/src/go/pt-galera-log-explainer/types/timeline.go
+++ b/src/go/pt-galera-log-explainer/types/timeline.go
@@ -40,11 +40,10 @@ func (timeline Timeline) MergeByIdentifier(lt LocalTimeline) {
 
 func (timeline Timeline) MergeByDirectory(path string, lt LocalTimeline) {
 	node := filepath.Base(filepath.Dir(path))
-MERGE:
 	for _, lt2 := range timeline {
 		if len(lt2) > 0 && node == filepath.Base(filepath.Dir(lt2[0].LogCtx.FilePath)) {
 			lt = MergeTimeline(lt2, lt)
-			break MERGE
+			break
 		}
 	}
 	timeline[node] = lt
@@ -59,7 +58,6 @@ func (timeline Timeline) MergeByPodnameElsePath(path string, lt LocalTimeline) {
 		timeline[path] = lt
 		return
 	}
-MERGE:
 	for _, lt2 := range timeline {
 		if len(lt2) == 0 {
 			continue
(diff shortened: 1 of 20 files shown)


svetasmirnova (approved):

src/go/pt-galera-log-explainer/types/timeline.go
```diff
@@ -40,14 +40,42 @@ func (timeline Timeline) MergeByIdentifier(lt LocalTimeline) {
 
 func (timeline Timeline) MergeByDirectory(path string, lt LocalTimeline) {
 	node := filepath.Base(filepath.Dir(path))
+MERGE:
```
artemgavrilov: Minor: It's redundant for non-nested loops

src/go/pt-galera-log-explainer/types/timeline.go
```diff
@@ -40,14 +40,42 @@ func (timeline Timeline) MergeByIdentifier(lt LocalTimeline) {
...
+// MergeByPodnameElsePath will try to keep podnames as identifier as it make more sense
+// if it does not have any metadata to use, it will resort to file paths
+// Merging operator logs is rare, but it happens when following multiple pt-k8s-debug-collector dumps
+func (timeline Timeline) MergeByPodnameElsePath(path string, lt LocalTimeline) {
+	metadata := lt[len(lt)-1].LogCtx.OperatorMetadata
+	if metadata == nil {
+		timeline[path] = lt
+		return
+	}
+MERGE:
```
artemgavrilov: Minor: same

artemgavrilov (approved):

src/go/pt-galera-log-explainer/types/timeline.go
ylacancellera: I thought it could help readability, but after a second take, I agree. Removed

PR #787, opened 2024-03-18
PT-2275 - "System Summary" PMM graph breaks when perl is not available

svetasmirnova:
- Added check for Perl binary into lib/bash/parse_options.sh
- Updated tools
- Updated t/lib/bash/parse_options.sh, so it includes required package alt_cmds.sh

- [ ] Documentation updated
- [x] Test suite update

### DIFF
--- a/bin/pt-ioprofile
+++ b/bin/pt-ioprofile
@@ -65,6 +65,7 @@ _d () {
 
 
 
+
 set -u
 
 ARGV=""           # Non-option args (probably input files)
@@ -93,6 +94,11 @@ usage_or_errors() {
    if [ "$OPT_VERSION" ]; then
       version=$(grep '^pt-[^ ]\+ [0-9]' "$file")
       echo "$version"
+      return 1
+   fi
+
+   if [ -z $(_which perl) ]; then
+      echo "Perl binary required to run this tool"
       return 1
    fi
 
--- a/bin/pt-mext
+++ b/bin/pt-mext
@@ -106,6 +106,7 @@ rm_tmpdir() {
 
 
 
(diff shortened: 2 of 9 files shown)


bin/pt-ioprofile
```diff
@@ -96,6 +97,11 @@ usage_or_errors() {
       return 1
    fi
 
+   if [ -z $(_which perl) ]; then
```
EvgeniyPatlan, line 100:
```suggestion
   if [ -z $(which perl) ]; then
```

bin/pt-mext
```diff
@@ -137,6 +138,11 @@ usage_or_errors() {
       return 1
    fi
 
+   if [ -z $(_which perl) ]; then
```
EvgeniyPatlan, line 141:
```suggestion
   if [ -z $(which perl) ]; then
```

EvgeniyPatlan (changes requested):
_which is in all files....

bin/pt-ioprofile
svetasmirnova: _which is the Toolkit custom function name, defined in lib/bash/alt_cmds.sh This is a wrapper around "regular" which with extra safety checks.

bin/pt-mext
svetasmirnova: _which is the Toolkit custom function name, defined in lib/bash/alt_cmds.sh This is a wrapper around "regular" which with extra safety checks.

BupycHuk (approved):

EvgeniyPatlan (approved):

PR #799, opened 2024-04-11
Added exceptions for advices and thr to .typos.toml

svetasmirnova:
- [ ] Documentation updated
- [ ] Test suite update

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -33,7 +33,6 @@ extend-ignore-re = [
 ]
 
 [default.extend-identifiers]
-"advices" = "advices" # typo in version_check.go
 "Bootstraping" = "Bootstraping" # typo in Galera lib
 "Continuent" = "Continuent"
 "END_ND_SUMMARY" = "END_ND_SUMMARY"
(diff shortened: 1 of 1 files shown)


.typos.toml
```diff
@@ -33,6 +33,7 @@ extend-ignore-re = [
 ]
 
 [default.extend-identifiers]
+"advices" = "advices" # typo in version_check.go
```
artemgavrilov: Why not to fix typo instead? Or we have there singular and plural `advise`?

artemgavrilov (approved):

.typos.toml
svetasmirnova: This is the variable name: array of multiple advice. But can change it to something like advs.

artemgavrilov: How about this? https://github.com/percona/percona-toolkit/pull/801

svetasmirnova: Looks good.

EvgeniyPatlan (approved):

.typos.toml
artemgavrilov: Merged

PR #811, opened 2024-04-29
Add flag for --client to provide path to kubectl

cshiv:
- [ ] Documentation updated
- [ ] Test suite update

This is useful for clients like openshift etc


svetasmirnova (changes requested):
Please add a test case.

PR #812, opened 2024-05-07
Add: github action libmodulesupdated

ylacancellera:

### DIFF
--- a/util/check_bin_md5sum
+++ b/util/check_bin_md5sum
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+for tool in bin/*; do 
+	shasum=$(md5sum $tool)
+	util/update-modules $tool; 
+	rc="$?"
+	if [ $? -ne 0 ]; then
+		echo "'util/update-modules $tool' exited with code $rc"
+	fi
+	
+	updated_shasum=$(md5sum $tool)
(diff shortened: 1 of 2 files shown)


svetasmirnova (approved):

util/check_bin_md5sum
```diff
@@ -0,0 +1,16 @@
...
+	rc="$?"
+	if [ $? -ne 0 ]; then
+		echo "'util/update-modules $tool' exited with code $rc"
+	fi
+	
+	updated_shasum=$(md5sum $tool)
+	if [ "$shasum" != "$updated_shasum" ]; then
+		echo "$tool had unsynced changes from lib, md5sum before and after running './util/update-modules $tool' is different"
+		exit 1
+	fi
```
EvgeniyPatlan, line 15: shouldn't we execute it for all tools and collect all possible failures?
So we can skip exit 1 and instead create an array of failed tools?

ylacancellera: We could yes, the idea was that if one tool was not updated, we should update-modules for all anyway

Would it help to have the full list ?

EvgeniyPatlan (approved):

util/check_bin_md5sum
EvgeniyPatlan: Let's keep it as is

PR #839, opened 2024-07-09
Update pt-mysql-summary

MASTERGOKUL:
Changed the deprecated variable name tx_isolation to transaction_isolation

"`transaction_isolation` was added in MySQL 5.7.20 as an alias for `tx_isolation`, which is now deprecated and is removed in MySQL 8.0. Applications should be adjusted to use `transaction_isolation` in preference to `tx_isolation`"

reference : https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_tx_isolation

### DIFF
--- a/lib/bash/report_mysql_info.sh
+++ b/lib/bash/report_mysql_info.sh
@@ -1018,13 +1018,9 @@ section_innodb () {
    # dynamically allocate variable name - transaction_isolation
    local mysql_version=$(get_var version "$variables_file" | awk -F'-' '{print $1}')
    local transaction_isolation_var="tx_isolation"
-   version_greater_equal() { # Function to compare versions
-      [ "$(printf '%s\n' "$1" "$2" | sort -V | head -n1)" = "$2" ]
-   }
-   if version_greater_equal "$mysql_version" "5.7.20"; then # true if version >= 5.7.20
+   if [ "$mysql_version" '>' "5.7.19" ]; then # true if version >= 5.7.20
          transaction_isolation_var="transaction_isolation"
    fi
-
 
    name_val "Buffer Pool Fill"   "$(fuzzy_pct ${bp_fill} ${bp_pags})"
    name_val "Buffer Pool Dirty"  "$(fuzzy_pct ${bp_dirt} ${bp_pags})"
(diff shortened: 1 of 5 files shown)


svetasmirnova:
Thank you for your contribution. You changed the auto-generated code. The fix should go into the lib/bash/report_mysql_info.sh file. Please fix, then run `./util/update-modules bin/pt-mysql-summary `, then run regression tests (I believe they also need to be adjusted).

Also, pt-mysql-summary should support older versions, so we need a condition here. Also, tests should be adjusted for 5.7.20+ and elder versions.

MASTERGOKUL:
hello @svetasmirnova  I updated the code with mysql version checks it is working fine as I run it. I am not aware of the regression tests, could you please infer me how to do it.

image
working fine actually 👆

lib/bash/report_mysql_info.sh
```diff
@@ -1014,6 +1014,18 @@ section_innodb () {
    local bp_free="$(get_var Innodb_buffer_pool_pages_free "$status_file")"
    local bp_dirt="$(get_var Innodb_buffer_pool_pages_dirty "$status_file")"
    local bp_fill=$((${bp_pags} - ${bp_free}))
+
+   # dynamically allocate variable name - transaction_isolation
+   local mysql_version=$(get_var version "$variables_file" | awk -F'-' '{print $1}')
+   local transaction_isolation_var="tx_isolation"
+   version_greater_equal() { # Function to compare versions
```
svetasmirnova: Why don't you compare version strings like we do on line 229 of `lib/bash/collect_mysql_info.sh`:
```if [ "$version" '<' "8.0" ]; then```

MASTERGOKUL: i thought that it could be good if we specify exact version like 5.7.20,  due to it is not a number we cant able to use `-ge`, is it ok if i use it like `if ["$mysql_version" '>' "5.7.19"] then`

svetasmirnova:
Thank you for the changes. I added a few more. Please review them and let us know if you agree.

MASTERGOKUL:
> Thank you for the changes. I added a few more. Please review them and let us know if you agree.

I agree with the changes, and thank you for adding it.

svetasmirnova (approved):

saikumar-vs (approved):

EvgeniyPatlan (approved):

PR #854, opened 2024-08-22
Option to Set Dynamic Minimum Chunk Size

ahirreddy:
- This PR introduces a flag that allows one to set the minimum/floor on the dynamic chunk size.
- This is important for use-cases in which a table has a insert rate that exceeds that which is allowed by the dynamically selected chunk size.

- [x] Documentation updated
- [ ] Test suite update

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -10318,10 +10318,11 @@ sub main {
                   $tbl->{nibble_time}, # is this amount of time
                );
 
-               # Enforce minimum chunk size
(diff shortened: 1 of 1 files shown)


svetasmirnova (changes requested):
Please add a test case. You can find examples in t/pt-online-schema-change

svetasmirnova:
Don't forget to add a test when the last chunk that copies the last few rows from the original table is smaller than the dynamic minimum chunk size.

PR #860, opened 2024-09-10
PT-2375 - fixed pt-table-sync for tables with generated columns

hpoettker:
The PR is intended to resolve this issue: https://perconadev.atlassian.net/browse/PT-2375

The tests added with `pt-2375.t` both fail with the current code base when executing the generated DML statements. With the proposed change, they run successfully. They test that both `REPLACE` and `UPDATE` statements are generated correctly by `pt-table-sync` when applied to tables with generated columns.

- [x] Documentation updated
- [x] Test suite update

### DIFF
--- a/bin/pt-table-sync
+++ b/bin/pt-table-sync
@@ -3708,6 +3708,7 @@ sub sort_cols {
    my @cols;
    if ( $self->{tbl_struct} ) {
       my $pos = $self->{tbl_struct}->{col_posn};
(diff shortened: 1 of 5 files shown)


hpoettker:
I updated the PR such that the change is also included in `lib/ChangeHandler.pm`. `pt-table-sync` is the only affected tool by this change.

hpoettker:

svetasmirnova (approved):

EvgeniyPatlan (approved):

PR #862, opened 2024-09-11
Fixed flaky tests

hpoettker:
The tests previously tested assertions against the replica after executing changes in the source without waiting for the replication to converge.

At least on my machine, the tests currently fail at least a couple of times when executed in loops like
```
for i in $(seq 10); do prove t/pt-table-sync/pt-1256.t; done
```
or
```
for i in $(seq 10); do prove t/pt-table-sync/char_chunking.t; done
```

- [x] Documentation updated
- [x] Test suite update


hpoettker:

EvgeniyPatlan (approved):

svetasmirnova (approved):

PR #856, opened 2024-08-28
PT-2322 - Fixing the pt-mysql-summary to detect jemalloc

Robertoh98:
Fixing the bug reported at https://perconadev.atlassian.net/browse/PT-2322.

On function report_jemalloc_enabled, the script couldn't get the right data due to a space on the grep done to gather the information, hence showing the jemalloc value empty.

PR #873, opened 2024-10-08
PT-2322 - Fixing the pt-mysql-summary to detect jemalloc #856

Robertoh98:
Fixing the bug reported at https://perconadev.atlassian.net/browse/PT-2322.

On function report_jemalloc_enabled, the script couldn't get the right data due to a space on the grep done to gather the information, hence showing the jemalloc value empty.


svetasmirnova (changes requested):
Why did you remove packages code from pt-mysql-summary?

Please also add a test case. You will find examples in t/pt-mysql-summary and t/lib/bash/report_mysql_info.t

svetasmirnova:
We will create new fix for this.

PR #876, opened 2024-10-16
Update pt-kill  Added new features that can support tidb

suyestyle:
- [x] Documentation updated
- [x] Test suite update

### DIFF
--- a/bin/pt-kill
+++ b/bin/pt-kill
@@ -4,6 +4,7 @@
 # See "COPYRIGHT, LICENSE, AND WARRANTY" at the end of this file for legal
 # notices and disclaimers.
 
+use DBI;
 use strict;
 use warnings FATAL => 'all';
 
@@ -7131,11 +7132,17 @@ sub main {
    my $kill;          # callback to KILL
    my $kill_sth;
    my $kill_sql;
+   my $db_type = $o->get('db-type') ;
+
    if ( $o->get('rds') ){
        $kill_sql = $o->get('kill-query') ? 'CALL mysql.rds_kill_query(?)' : 'CALL mysql.rds_kill(?)';
    }
    else{
        $kill_sql = $o->get('kill-query') ? 'KILL QUERY ?' : 'KILL ?';
+   }
+
+   if (defined $db_type && length $db_type > 0 && $db_type eq "tidb") {
+      $kill_sql = $o->get('kill-query') ? 'KILL QUERY ?' : 'KILL tidb ?';
    }
    my $files;
    if ( $files = $o->get('test-matching') ) {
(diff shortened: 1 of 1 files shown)


svetasmirnova (changes requested):
Please add a test case and make "mysql" default value for the "db-type" parameter. You can find an example of how to do it in the definition of option recursion-method for pt-online-schema-change utility.

suyestyle:
If you want to add rules to mysql clusters, you can do so
```
pt-kill --log-dsn D=3306_kill_log,t=99755_r_3306_killsql_log,h=192.168.1.1,P=3307,u=dba,p=password --host=192.168.1.2 --port=3306 --user=dba --password=password --create-log-table --no-version-check  --busy-time=4s --match-command=Query --match-info='SELECT|select|Select' --ignore-user=inception --victim all --interval 1 --kill --db-type=mysql --daemonize
```
Or do this
```
pt-kill --log-dsn D=3306_kill_log,t=99755_r_3306_killsql_log,h=192.168.1.1,P=3307,u=dba,p=password --host=192.168.1.2 --port=3306 --user=dba --password=password --create-log-table --no-version-check  --busy-time=4s --match-command=Query --match-info='SELECT|select|Select' --ignore-user=inception --victim all --interval 1 --kill --daemonize
```


If you want to add rules to tidb clusters, you can do so
```
pt-kill --log-dsn D=3306_kill_log,t=99755_r_3306_killsql_log,h=192.168.1.1,P=3307,u=dba,p=password --host=192.168.1.2 --port=4001 --user=dba --password=password --create-log-table --no-version-check  --busy-time=4s --match-command=Query --match-info='SELECT|select|Select' --ignore-user=inception --victim all --interval 1 --kill --db-type=tidb --daemonize
```

bin/pt-kill
```diff
@@ -4,6 +4,7 @@
 # See "COPYRIGHT, LICENSE, AND WARRANTY" at the end of this file for legal
 # notices and disclaimers.
 
+use DBI;
```
svetasmirnova, line 7: This is auto-generated code. No change should be here.

I also do not understand why you need to explicitly import DBI for the code you propose.

bin/pt-kill
```diff
@@ -8011,6 +8030,11 @@ pt-kill does not provide any safeguards so code carefully!
 
 It is permissible for the code to have side effects (to alter C<$event>).
 
+=item --db-type
+
+type: string
+
+mysql; tidb
```
svetasmirnova, line 8037: Here should be the default value for mysql. Please check other options where the default value is defined.

svetasmirnova (changes requested):
Please resolve my comments and add a test case. You will find examples in t/pt-kill.

suyestyle:
How do I provide the use cases you're talking about.

svetasmirnova:
@suyestyle Please check CONTRIBUTING.md and files in t/pt-kill.

suyestyle:
Thank you for your patience.

I have put my use case in the above answer, is there any problem with this way?

I can't find any use case information here on t/pt-kill, a lot of it is code information.

svetasmirnova:
@suyestyle I am not asking about the use case. I am asking about regression tests (https://en.wikipedia.org/wiki/Regression_testing) that we run before releasing the new version of Percona Toolkit to ensure that no previously added feature is broken. You will find examples of such tests in the t/pt-kill directory. There is also information at https://github.com/percona/percona-toolkit/blob/3.x/CONTRIBUTING.md#setting-up-the-development-environment-and-testing

suyestyle:
The first contact with open source project is not familiar with the operation process, it seems to me a little complicated, then ignore this PR, thank you.

PR #890, opened 2024-12-04
pt-config-diff: MySQL truncates run-time variable values longer than 1024 characters

dbnski:
`pt-config-diff` relies on `SHOW GLOBAL VARIABLES` for listing run-time variables. MySQL 5.7 and newer implement the command by reading the contents of the `performance_schema` table called `global_variables`, but the table definition uses `VARCHAR(1024)` for the value field. Any value longer than 1024 characters becomes truncated both in the table as well as in the `SHOW GLOBAL VARIABLES` command output. This makes comparing long run-time variables unreliable. The patch works around the problem by reading full values from the `@@global` object whenever necessary.


svetasmirnova (changes requested):
You modified auto-generated code. Please place your changes into lib/MySQLConfig.pm, then run ./util/update-modules command.

Please also add a test case. You will find examples in t/lib/MySQLConfig.t and in t/pt-config-diff/

dbnski:
Updated MySQLConfig.pm and added test cases.

svetasmirnova:
Thank you for the patch update and the test case. I modified the test variable so it works with versions 5.7, 8.0, and 8.4. Please review and let me know if it is OK to remove long_vars_57_1.cnf, long_vars_57_2.cnf, long_vars_80_1.cnf, long_vars_80_2.cnf samples.

EvgeniyPatlan (approved):

dbnski:
@svetasmirnova Sure, that is great. I was looking for a var that'd work across all versions, but didn't catch that one.

svetasmirnova (approved):

PR #891, opened 2024-12-06
PT-2401 - remove check to fix the 'table does not exist' problem when lower_case_table_names=2 is set

ehooi:
pt-online-schema-change failed with 'table `TableName` does not exist', but the table exists.

It seems that #516 introduces this issue.

The variable $tbl changed to lowercase as `tablename` but MySQL returned the original case `TableName` when `lower_case_table_names=2` is set.
As a result, the check that I removed fails.

Since the check appeared to be primarily preventive, I decided to remove it.


svetasmirnova (changes requested):
Please add a test case. You can take t/pt-online-schema-change/case-insensitive.t as example.

svetasmirnova (approved):

svetasmirnova:
I modified the code so it still checks the table name but now takes into account the value of the lower_case_table_names option. I hope this is OK for you. Regarding the test case, I think t/pt-online-schema-change/case-insensitive.t is enough because we cannot modify this test to check the lower_case_table_names option on Linux anyway.

EvgeniyPatlan (approved):

PR #903, opened 2024-12-21
PT-2410 - Fixing the export with output-format=csv when there are null values

Robertoh98:
Fixing the bug reported at https://perconadev.atlassian.net/browse/PT-2410.

I added to the if on the escape function the \N to be escaped from the enclosed by ".


svetasmirnova:
Thank you for the fix. Please add a test case for your fix.

svetasmirnova (changes requested):
Please add a test case. You will find examples in directory t/pt-archiver

svetasmirnova (approved):

EvgeniyPatlan (approved):

PR #919, opened 2025-01-28
PT-2389 Fix unconditional ALTER on new table with resume

drwonky:
- [x] Documentation updated
- [ ] Test suite update

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -9912,7 +9912,7 @@ sub main {
       );
    }
 
-   if ( my $alter = $o->get('alter') ) {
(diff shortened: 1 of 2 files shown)


drwonky:
I'm having trouble running the test suite when following the directions in the CONTRIBUTING.md precisely.  I am getting an SSL error from the PERL DBI library: `SSL connection error: Enforcing SSL encryption is not supported`.  I noted that the resume test for PT-1717 did a null alter and that's why PT-2389 was not surfaced, so I updated that test to add a column instead of doing a null alter.

svetasmirnova (approved):

EvgeniyPatlan (approved):

PR #944, opened 2025-03-26
PT-2407 - pt-online-schema-change exit status(return code) is 0 even …

svetasmirnova:
…if it does NOT succeed

- Changed die call for the case when ALTER TABLE fails for the new table
- Removed stalled comment from Frank
- Added test case

- [ ] Documentation updated
- [x] Test suite update

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -9608,13 +9608,13 @@ sub main {
          #   doesn't match '(?-xism:Failed to find a unique new table name)'
 
          my $original_error = $EVAL_ERROR;
-my $original_error_code;
-if ( $? ) {
-   $original_error_code = $?;
-} 
-else {
-   $original_error_code = $!;
-}
+         my $original_error_code;
+         if ( $? ) {
+            $original_error_code = $?;
+         } 
+         else {
+            $original_error_code = $!;
+         }
 
          $SIG{__DIE__} = 'DEFAULT';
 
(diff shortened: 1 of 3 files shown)


bin/pt-online-schema-change
```diff
@@ -9607,11 +9607,10 @@ sub main {
          #                   ''
          #   doesn't match '(?-xism:Failed to find a unique new table name)'
 
-         # (*) Frank:  commented them out because it caused infinite loop
-         # and the mentioned test error doesn't arise
-
          my $original_error = $EVAL_ERROR;
-         my $original_error_code = $?;
+         my $original_error_code = $? ? $? : $!;
```
EvgeniyPatlan: ternary operators are not recommended according to Perl Best Practices so it might be better to replace it with something like this

```
my $original_error_code;
if ($?) {
    $original_error_code = $?;
} 
else {
    $original_error_code = $!;
}
```

EvgeniyPatlan (approved):

bin/pt-online-schema-change
svetasmirnova:
```suggestion
my $original_error_code;
if ( $? ) {
   $original_error_code = $?;
} 
else {
   $original_error_code = $!;
}
```

bin/pt-online-schema-change
```diff
@@ -9607,11 +9607,16 @@ sub main {
...
-         my $original_error_code = $?;
+my $original_error_code;
+if ( $? ) {
+   $original_error_code = $?;
+} 
+else {
+   $original_error_code = $!;
+}
+
+         $SIG{__DIE__} = 'DEFAULT';
```
svetasmirnova:
```suggestion
         my $original_error_code;
         if ( $? ) {
            $original_error_code = $?;
         } 
         else {
            $original_error_code = $!;
         }

         $SIG{__DIE__} = 'DEFAULT';
```

saikumar-vs (approved):

PR #965, opened 2025-05-28
PT-2451 Added missing case-insensitivity flags to m/maria/ regexps

init-penguin:
Kinda trivial bugfix: pt-* tools sometimes cannot differentiate between MariaDB/MySQL because of a missing `/i` at the end of `m/mariadb/`.

- [+] The contributed code is licensed under GPL v2.0
- [+] Contributor Licence Agreement (CLA) is signed
- [-] util/update-modules has been ran
- [-] Documentation updated
- [-] Test suite update

No changes in libs were made thus no update-modules.
No changes in docs or tests are required in my opinion..

### DIFF
--- a/bin/pt-archiver
+++ b/bin/pt-archiver
@@ -3926,7 +3926,7 @@ sub _find_replicas_by_hosts {
    my $vp = VersionParser->new($dbh);
    my $sql = 'SHOW REPLICAS';
    my $source_name = 'source';
-   if ( $vp lt '8.1' || $vp->flavor() =~ m/maria/ ) {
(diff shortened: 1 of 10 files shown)


svetasmirnova:
Thank you for the contribution. You updated auto-generated code. Please put your changes into lib/MasterSlave.pm, then run util/update-modules on files in the bin directory. Also, this change should be in other places where the version is checked.

svetasmirnova:
Btw, we also have PR-981 with the same issues as yours, but with a test case.

svetasmirnova:
We will merge https://github.com/percona/percona-toolkit/pull/981

PR #966, opened 2025-05-30
PT-2453: Add -skip-pod-summary for pt-k8s-debug-collector

imp3142:
In some cases, collecting additional data from pods is unnecessary and, due to the current tool design, can cause significant delays. This commit introduces a workaround to skip such collection when it's not required.


svetasmirnova:
Please add a test case.

svetasmirnova (approved):

artemgavrilov (approved):

PR #976, opened 2025-07-02
PT-2418: Fix column data loss during RENAME COLUMN operations

marek-knappe:
The pt-online-schema-change tool was not properly handling RENAME COLUMN syntax (MySQL 8.0+), causing renamed columns to be excluded from the data copy operation. This resulted in NULL values for renamed columns after the schema change.

The issue was in the find_renamed_cols() function which only supported CHANGE COLUMN syntax but not RENAME COLUMN syntax. Added support for RENAME COLUMN parsing to properly detect column renames and include them in the common_cols list for data copying.

Fixes: PT-2418

- [X] util/update-modules has been ran
- [X] Documentation updated
- [X] Test suite update

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -11239,7 +11239,7 @@ sub find_renamed_cols {
       next if lc($orig_tbl) eq lc($new_tbl);
       $renames{lc($orig_tbl)} = $new_tbl;
    }
-   
+
    # Also handle RENAME COLUMN syntax (MySQL 8.0+)
    while ( $alter =~ /$alter_rename_col_re/g ) {
       my ($orig, $new) = map { $tp->ansi_to_legacy($_) } $1, $2;
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -11226,6 +11226,8 @@ sub find_renamed_cols {
 
    my $alter_change_col_re = qr/\bCHANGE \s+ (?:COLUMN \s+)?
                                 ($table_ident) \s+ ($table_ident)/ix;
+   my $alter_rename_col_re = qr/\bRENAME \s+ COLUMN \s+
+                                ($table_ident) \s+ TO \s+ ($table_ident)/ix;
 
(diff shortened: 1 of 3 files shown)


svetasmirnova (approved):

svetasmirnova:
Thank you for the contribution. I added cleanup for the test and additional checks. Please let us know if you disagree.

bin/pt-online-schema-change
```diff
@@ -11237,6 +11239,17 @@ sub find_renamed_cols {
...
+
+   # Also handle RENAME COLUMN syntax (MySQL 8.0+)
+   while ( $alter =~ /$alter_rename_col_re/g ) {
+      my ($orig, $new) = map { $tp->ansi_to_legacy($_) } $1, $2;
+      next unless $orig && $new;
+      my (undef, $orig_tbl) = Quoter->split_unquote($orig);
+      my (undef, $new_tbl)  = Quoter->split_unquote($new);
+      # Silly but plausible: RENAME COLUMN same_name TO same_name ...
+      next if lc($orig_tbl) eq lc($new_tbl);
+      $renames{lc($orig_tbl)} = $new_tbl;
```
EvgeniyPatlan, line 11251: I can only suggest an improvement here. We are doing lower case several times. 
It might be possible to introduce new variables and work with them. Here is the example of the change:

```
my $orig_tbl_lc = lc($orig_tbl);
my $new_tbl_lc  = lc($new_tbl);
next if $orig_tbl_lc eq $new_tbl_lc;
$renames{$orig_tbl_lc} = $new_tbl;
```

EvgeniyPatlan (approved):

bin/pt-online-schema-change
marek-knappe: I just followed the same convention used in the block above.
I don’t think introducing new variables is necessary here — it doesn’t save any real CPU and adds extra lines that make the code slightly harder to read, but if that's the wish of owners I'm happy to change it :)

PR #1003, opened 2025-08-28
PT-2305 - pt-online-schema-change should error if server is a slave in row based replication

svetasmirnova:
- Implemented fix and test case
- Updated documentation

- [x] Documentation updated
- [ ] Test suite update

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -9202,7 +9202,6 @@ sub main {
 
             if ( $is_source_of) {
                my $source_binlog_format = $source_cxn->dbh()->selectrow_arrayref("SHOW GLOBAL VARIABLES LIKE 'binlog_format'");
-               use Data::Dumper;
                if ( uc @$source_binlog_format[1] ne 'STATEMENT' ) {
                   _die("Server " . $dp->as_string($cxn->dsn())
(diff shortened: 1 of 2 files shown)


bin/pt-online-schema-change
```diff
@@ -9181,6 +9181,42 @@ sub main {
...
+            if ( $EVAL_ERROR ) {
+               $EVAL_ERROR =~ m/The replica is connected to (\d+) but the source's port is \d+/;
+               if ( !$1 || $1 != $source->{P} ) {
+                  $is_source_of = 0;
+               }
+            }
+
+            if ( $is_source_of) {
+               my $source_binlog_format = $source_cxn->dbh()->selectrow_arrayref("SHOW GLOBAL VARIABLES LIKE 'binlog_format'");
+               use Data::Dumper;
```
EvgeniyPatlan: we have this use at the beginning of the script but we load only qw(). So better to update it at the beginning

EvgeniyPatlan (changes requested):

bin/pt-online-schema-change
svetasmirnova: Oh, this `use Data::Dumper` should not be here at all, removed.

EvgeniyPatlan (approved):

saikumar-vs (approved):

PR #1005, opened 2025-09-02
Pt 2289 allow pt stalk do disable ps lock transactions data collection via parameter

svetasmirnova:
- [x] Documentation updated
- [x] Test suite update

### DIFF
--- a/bin/pt-stalk
+++ b/bin/pt-stalk
@@ -971,7 +971,7 @@ collect_mysql_data_one() {
       fi
    fi
 
-   if ! [ "${OPT_SKIP_COLLECTION}" =~ "mysqladmin" ]; then
+   if ! [[ "${OPT_SKIP_COLLECTION[@]}" =~ "mysqladmin" ]]; then
       $CMD_MYSQLADMIN $EXT_ARGV ext -i$OPT_SLEEP_COLLECT -c$cnt >>"$d/$p-mysqladmin" &
       mysqladmin_pid=$!
    fi
@@ -1016,7 +1016,7 @@ collect_system_data() {
 
 collect_mysql_data_loop() {
 
-   if ! [ "${OPT_SKIP_COLLECTION}" =~ "processlist" ]; then
+   if ! [[ "${OPT_SKIP_COLLECTION[@]}" =~ "processlist" ]]; then
       (echo $ts; $CMD_MYSQL $EXT_ARGV -e "SHOW FULL PROCESSLIST\G") \
       >> "$d/$p-processlist" &
    fi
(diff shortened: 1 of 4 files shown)


bin/pt-stalk
```diff
@@ -971,8 +971,10 @@ collect_mysql_data_one() {
       fi
    fi
 
-   $CMD_MYSQLADMIN $EXT_ARGV ext -i$OPT_SLEEP_COLLECT -c$cnt >>"$d/$p-mysqladmin" &
-   mysqladmin_pid=$!
+   if ! [[ "${OPT_SKIP_COLLECTION[@]}" =~ "mysqladmin" ]]; then
```
EvgeniyPatlan: Bash does not support regex matching directly against arrays using =~.

use something like this
```
should_skip=false
for item in "${OPT_SKIP_COLLECTION[@]}"; do
    if [[ "$item" == "mysqladmin" ]]; then
        should_skip=true
        break
    fi
done

if ! $should_skip; then
    "$CMD_MYSQLADMIN" $EXT_ARGV ext -i"$OPT_SLEEP_COLLECT" -c"$cnt" >> "$d/$p-mysqladmin" &
    mysqladmin_pid=$!
fi
```

bin/pt-stalk
```diff
@@ -1014,18 +1016,25 @@ collect_system_data() {
 
 collect_mysql_data_loop() {
 
-
+   if ! [[ "${OPT_SKIP_COLLECTION[@]}" =~ "processlist" ]]; then
```
EvgeniyPatlan: the same

bin/pt-stalk
```diff
@@ -1014,18 +1016,25 @@ collect_system_data() {
...
-      >> "$d/$p-threads" &
+   fi
+
+   (echo $ts; $CMD_MYSQL $EXT_ARGV -e "SELECT * FROM performance_schema.threads\G") \
+   >> "$d/$p-threads" &
 
    if [ "$have_lock_waits_table" ]; then
-      (echo $ts; lock_waits "$d/lock_waits.running")   >>"$d/$p-lock-waits" &
-      (echo $ts; transactions) >>"$d/$p-transactions" &
+      if ! [[ "${OPT_SKIP_COLLECTION[@]}" =~ "lock-waits" ]]; then
```
EvgeniyPatlan: the same

EvgeniyPatlan (changes requested):

bin/pt-stalk
svetasmirnova: Fixed, please review again

bin/pt-stalk
```diff
@@ -1356,18 +1367,33 @@ collect_mysql_variables() {
...
+      $CMD_MYSQL $EXT_ARGV -e "$sql" >> $outfile
+   fi
+
+}
+
+_should_skip() {
+   local name=$1
+
+   for item in "${OPT_SKIP_COLLECTION[@]}"; do
+   echo $item >> /tmp/sveta
```
EvgeniyPatlan: let's implement another name here

EvgeniyPatlan (changes requested):

bin/pt-stalk
svetasmirnova: OMG, I need some automation to remove those print debug messages.

EvgeniyPatlan (approved):

saikumar-vs (approved):

PR #1013, opened 2025-09-09
PMM-11406 Better check to prevent errors from missing fields.

JiriCtvrtka:
- [ ] Documentation updated
- [x] Test suite update

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -11668,7 +11668,7 @@ sub check_orig_table {
          Cxn          => $cxn,
          tbl          => $orig_tbl,
          chunk_size   => $o->get('chunk-size'),
-         chunk_indx   => $o->get('chunk-index'), # typos:ignore-line
+         chunk_index  => $o->get('chunk-index'),
          OptionParser => $o,
          TableParser  => $tp,
       );
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -11668,7 +11668,7 @@ sub check_orig_table {
          Cxn          => $cxn,
          tbl          => $orig_tbl,
          chunk_size   => $o->get('chunk-size'),
-         chunk_indx   => $o->get('chunk-index'),
+         chunk_indx   => $o->get('chunk-index'), # typos:ignore-line
          OptionParser => $o,
          TableParser  => $tp,
       );
@@ -13512,7 +13512,7 @@ New rows will use the user defined default value if specified for the column.
 =item --only-same-schema-fks
 
 Check foreigns keys only on tables on the same schema than the original table.
-This option is dangerous since if you have FKs refenrencing tables in other
+This option is dangerous since if you have FKs referencing tables in other
 schemas, they won't be detected.
 
 
--- a/src/go/pt-mongodb-summary/main_test.go
+++ b/src/go/pt-mongodb-summary/main_test.go
@@ -58,9 +58,7 @@ func TestGetHostInfoResult(t *testing.T) {
 
 	host, err := getHostInfo(ctx, client)
 	require.NoError(t, err, "getHostInfo error")
-
-	notEmpty := host.ProcessName != "" || host.Version != "" || host.ProcPath != "" || host.ProcUserName != "" || !host.ProcCreateTime.IsZero()
-	require.True(t, notEmpty, "At least one field in host should not be empty to ensure we got some data")
+	require.NotEmpty(t, host)
 }
 
 func TestClusterWideInfo(t *testing.T) {
--- a/src/go/pt-mongodb-summary/main_test.go
+++ b/src/go/pt-mongodb-summary/main_test.go
@@ -10,7 +10,6 @@ import (
 	"github.com/pborman/getopt"
 
 	tu "github.com/percona/percona-toolkit/src/go/internal/testutils"
-	"github.com/percona/percona-toolkit/src/go/mongolib/proto"
 )
 
 func TestGetHostInfo(t *testing.T) {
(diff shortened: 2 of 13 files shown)


svetasmirnova (changes requested):
Please add a test case.

svetasmirnova:
I added a fix for the logrus.Infof error that broke tests for main. Now they are passed. I hope this is OK.

JiriCtvrtka:
@svetasmirnova Currently, we are not checking the output from the getHostInfo method at all. I’ve made a small change in the logic to make it more stable and robust for setup where getCmdLineOpts.recordStats or serverStatus.recordStats are missing. Right now, this situation ends with an error and produces no output at all. With this change, the missing fields will simply be skipped, and the rest of the output will still be printed.

In the PMM team, our QA setup uses a sharded cluster with replica sets, but those two fields are always missing regardless of version or distribution.

I can create a new test like this:
```
func TestGetHostInfo(t *testing.T) {
    assert := require.New(t)

    ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
    defer cancel()

    client, err := tu.TestClient(ctx, tu.MongoDBShard1PrimaryPort)
    assert.NoError(err, "cannot get a new MongoDB client")

    host, err := getHostInfo(ctx, client)
    assert.NoError(err, "getHostInfo error")

    assert.NotEmpty(host.ProcessName, "ProcessName should not be empty if serverStatus succeeds")
    assert.NotEmpty(host.Version, "Version should not be empty if serverStatus succeeds")
    assert.NotEmpty(host.ProcPath, "ProcPath should not be empty if getProcInfo succeeds")
    assert.NotEmpty(host.ProcUserName, "ProcUserName should not be empty if getProcInfo succeeds")
    assert.False(host.ProcCreateTime.IsZero(), "ProcCreateTime should not be zero if getProcInfo succeeds")
}
```
This would cover those cases, but since the Percona Toolkit test environment always has these fields, it won’t be able to test the missing-field scenario anyway. 

Thanks for fix of print. Also in scope of PR I removed unused method in main_test file.

BupycHuk (approved):

svetasmirnova:
@JiriCtvrtka Yes, please add this test still. I have plans for improving the Percona Toolkit test environment btw. Not sure when find time for it.

EvgeniyPatlan (approved):

JiriCtvrtka:
Test added.

svetasmirnova:
@JiriCtvrtka 

test fails with:
```
$ go test -v
=== RUN   TestGetHostInfo
=== RUN   TestGetHostInfo/from_mongos
=== RUN   TestGetHostInfo/from_mongod
--- PASS: TestGetHostInfo (0.14s)
    --- PASS: TestGetHostInfo/from_mongos (0.08s)
    --- PASS: TestGetHostInfo/from_mongod (0.06s)
=== RUN   TestGetHostInfoResult
    main_test.go:67: 
        	Error Trace:	/home/sveta/src/percona/percona-toolkit/src/go/pt-mongodb-summary/main_test.go:67
        	Error:      	Should NOT be empty, but was 
        	Test:       	TestGetHostInfoResult
        	Messages:   	ProcPath should not be empty if getProcInfo succeeds
--- FAIL: TestGetHostInfoResult (0.06s)
=== RUN   TestClusterWideInfo
=== RUN   TestClusterWideInfo/from_mongos
=== RUN   TestClusterWideInfo/from_mongod
    main_test.go:102: getClisterWideInfo error: getClusterwideInfo.listDatabases : (NotPrimaryOrSecondary) node is not in primary or recovering state
--- FAIL: TestClusterWideInfo (0.03s)
    --- PASS: TestClusterWideInfo/from_mongos (0.03s)
    --- FAIL: TestClusterWideInfo/from_mongod (0.00s)
=== RUN   TestParseArgs
--- PASS: TestParseArgs (0.00s)
FAIL
exit status 1
FAIL	github.com/percona/percona-toolkit/src/go/pt-mongodb-summary	0.236s
```
Is this expected?

src/go/pt-mongodb-summary/main_test.go
```diff
@@ -49,6 +49,26 @@ func TestGetHostInfo(t *testing.T) {
 	}
 }
 
+func TestGetHostInfoResult(t *testing.T) {
+	assert := require.New(t)
```
idoqo: not sure what a good name here is, but changing assert -> require might be misleading (since they're different behaviors though)

bin/pt-online-schema-change
```diff
@@ -11668,7 +11668,7 @@ sub check_orig_table {
          Cxn          => $cxn,
          tbl          => $orig_tbl,
          chunk_size   => $o->get('chunk-size'),
-         chunk_indx   => $o->get('chunk-index'),
+         chunk_index  => $o->get('chunk-index'),
```
JiriCtvrtka, line 11671: I think it does not work due to typo. Now it should be fixed.

JiriCtvrtka:
@svetasmirnova 
**TestGetHostInfoResult** – I rewrote the test to be as universal as possible and not dependent on configuration. If we have some data in the response, that’s OK. This still covers the original purpose of the change.

Regarding failing **TestClusterWideInfo/from_mongod** – that should not be related to my changes in this PR.

As for satisfying spelling/typos – I fixed them across the whole repository. One possible bug here because of this: https://github.com/percona/percona-toolkit/pull/1013/files#r2398484718

src/go/pt-mongodb-summary/main_test.go
```diff
@@ -85,16 +97,6 @@ func TestClusterWideInfo(t *testing.T) {
 	}
 }
 
-func addToCounters(ss proto.ServerStatus, increment int64) proto.ServerStatus {
```
JiriCtvrtka, line 88: Not used

JiriCtvrtka:
@svetasmirnova could you check and merge? See my comment: https://github.com/percona/percona-toolkit/pull/1013#issuecomment-3360904980. Thank you.

ademidoff:
> @svetasmirnova could you check and merge? See my comment: [#1013 (comment)](https://github.com/percona/percona-toolkit/pull/1013#issuecomment-3360904980). Thank you.

@svetasmirnova PTAL

svetasmirnova (approved):

PR #1018, opened 2025-10-17
Fix replica tracking and undefined DSN value warnings in pt-online-schema-change

zxzharmlesszxz:
…-schema-change

Avoids "Use of uninitialized value in string eq" warnings in ReplicaLagWaiter by safely handling undefined DSN host/port values. This change uses the defined-or operator (// '') to ensure comparisons work even when DSN fields are not initialized, preventing unnecessary reconnections to replicas.

- [ ] Documentation updated
- [ ] Test suite update

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -4377,10 +4377,10 @@ sub recurse_to_replicas {
 
    DBH: {
       if ( !defined $dbh ) {
-         foreach my $known_slave ( @{$args->{slaves}} ) {
-            if (($known_slave->{dsn}->{h} // '') eq ($slave_dsn->{h} // '') and
-                ($known_slave->{dsn}->{P} // '') eq ($slave_dsn->{P} // '')) {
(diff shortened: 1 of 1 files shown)


zxzharmlesszxz:
without this changes replica lag working only when servers on host table

svetasmirnova (changes requested):
You modified auto-generated code and you changes will be overwritten next time we will update modules for the Percona Toolkit. Please check CONTRIBUTING.md for instructions on how to modify shared code and write tests, required for all contributions.

svetasmirnova:
It is dangerous to allow having empty host or port for replicas DSN. If this situation happens when replicas are discovered automatically, this is a bug, and we need a repeatable test case reported at jira.percona.com. If a user uses the DSN table and specifies an empty host or port value, this is the user's fault, and we would do nothing about it. The error message could be nicer, but we cannot check what the user inserts into the DSN table before the tool runs.

PR #1022, opened 2025-10-29
add mysql_optional as dsn parameter

Thomblin:
This pull requests adds `mysql_ssl_option` as an dsn paramter to pt-online-schema-change.

Why?

Upgrading debian from bullseye to bookworm or trixie changes the behaviour of perl DBI. As a consequence perl DBI is not able to use ssl anymore when enforced (pt-online-schema-change raises the exception "SSL connection error: Enforcing SSL encryption is not supported"). 

See: [Debian Bug Report 1032074](https://bugs-devel.debian.org/cgi-bin/bugreport.cgi?bug=1032074)

I found no fix for this (using official sources). The only workaround to make it work is calling pt-online-schema-change with both `mysql_ssl=1` and `mysl_ssl_optional=1`

Please advice how and if tests or documentation need to be adjusted.

- [X] Individual Contributor License Agreement is signed
- [ ] Documentation updated
- [ ] Test suite update

### DIFF
--- a/bin/pt-archiver
+++ b/bin/pt-archiver
@@ -8121,6 +8121,12 @@ short form: -s; type: int
 
 Create SSL MySQL connection.
 
+=item --mysql_ssl_optional
+
+short form: -o; type: int
+
+Disables strict SSL enforcement and makes SSL connection optional.
+
 =item --no-ascend
 
 Do not use ascending index optimization.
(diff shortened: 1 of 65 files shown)


svetasmirnova:
Immediate workaround is to install DBD::mysql, depending on upstream or Percona libmysql, available in Percona repositories.

I will check proposed solution, so Percona Toolkit can run with default DBD::mysql, linked to libmariadb3 too. Currently I can only say that you modified auto-generated code that will be rewritten once we create packages or simply update modules. Check CONTRIBUTING.md for how to update modules and create tests for your contributions.

Thomblin:
@svetasmirnova I added the option to lib and tests

EvgeniyPatlan (approved):

svetasmirnova (approved):

PR #1023, opened 2025-10-30
Replace validation logic for data-dir in pt-osc to use innodb_directories.

peppla:
- [ ] Documentation updated
- [ ] Test suite update

Current validation checks that the directory exists and is writable by the user running pt-osc. This does not make sense because the tool can be executed remotely, and even locally, access is not required for the user running the tool. What is required is that the directory exists in `innodb_directories`.

This change removes the old validation and checks if the directory passed in data-dir exists.

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -9838,6 +9838,7 @@ sub main {
             Quoter           => $q,
             OptionParser     => $o,
             TableParser      => $tp,
+            server_version   => $server_version,
          );
       };
       if ( $EVAL_ERROR ) {
@@ -11372,6 +11373,7 @@ sub create_new_table {
    }
    my ($new_table_name, $orig_tbl, $cxn, $q, $o, $tp) = @args{@required_args};
    my $new_table_prefix = $args{new_table_prefix};
+   my $server_version = $args{server_version};
 
    # Get the original table struct.
    my $ddl = $tp->get_create_table(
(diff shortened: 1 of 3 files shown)


svetasmirnova (changes requested):
Thank you for the contribution. Please add a test case as described in CONTRIBUTING.md

svetasmirnova:
It is also possible to create such tables in innodb_data_home_dir.

svetasmirnova:
@peppla I added test case myself and also added a check in case if the directory is innodb_data_home_dir (also allowed). Please check changes and let me know if you disagree with them.

svetasmirnova (approved):

peppla:
I agree, of course. Thank you very much!

bin/pt-online-schema-change
```diff
@@ -11438,13 +11440,37 @@ sub create_new_table {
...
+         }
+
+         my @dirs = map {
+            my $dir = $_;
+            $dir =~ s{(?<!^)/\z}{};
+            $dir;
+         } split(/\s*;\s*/, $innodb_dirs);
+         my %dirs_lookup = map { $_ => 1 } @dirs;
+         my $data_dir = $o->get('data-dir');
+         $data_dir =~ s{(?<!^)/\z}{};
```
saikumar-vs: This can be removed,  trailing-slash regex applied again.

saikumar-vs (changes requested):

saikumar-vs (approved):

PR #1051, opened 2026-01-06
PT-2237 - fixed PostgreSQL DB logs collection

BON4:
### This fix ensures `pt-k8s-debug-collector` collects PostgreSQL database logs.

**What was added:**

* To keep changes minimal, the existing test `TestIndividualFiles` was modified to cover the new case.
  The test was refactored because the previous implementation couldn't handle files with variable names.
  In `/pgdata/<cluster_name>/pg_log`, logs are named according to the day/week, so filenames are not known at runtime.

* A new function `getAllFilesFromDirectory` was added to `dumper.go`.
  Its purpose is the same as the existing `getIndividualFiles`, but it supports extracting **multiple files** from a directory. It is useful because pg log filenames can vary, and there can be multiple log files.

(`/lib` has not been changed)
- [X] Documentation updated
- [X] Test suite update


src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -126,8 +129,13 @@ func New(location, namespace, resource string, kubeconfig string, forwardport st
 	}
 	sslSecrets := make([]sslSecret, 0)
 	filePaths := make([]string, 0)
+	dirPaths := make([]string, 0)
 	switch resourceType(resource) {
 	case "pg":
+		dirPaths = append(dirPaths,
+			"$PGBACKREST_DB_PATH/pg_log",
```
svetasmirnova: Where does the variable `PGBACKREST_DB_PATH` come from? It may not be set on the client machine where `pt-k8s-debug-collector` is running. Or could be set to a completely irrelevant value regarding the cluster it collects the data from.

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -497,6 +518,95 @@ func (d *Dumper) getIndividualFiles(namespace string, podName, path, location st
...
+		"-n", namespace, "-c", d.fileContainer, "exec", podName, "--",
+		"sh", "-c", fmt.Sprintf("echo %s", env),
+	}
+
+	fmt.Printf("DBG: running parseEnvs: %v\n", args)
+	out, err := d.runCmd(args...)
+	if err != nil {
+		return "", err
+	}
+	fmt.Printf("DBG: running parseEnvs output: %v\n", string(out))
```
svetasmirnova: Why do we need this if the tool uses log.Print, log.logError, and so on?

svetasmirnova: Also, debugging output should not be printed all the time.

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -561,27 +671,27 @@ func (d *Dumper) getPodSummary(resource, podName, crName string, namespace strin
...
-		err := cmdPortFwd.Run()
-		if err != nil {
-			d.logError(err.Error(), "port-forward")
-		}
-	}()
+	argPortFwd := []string{"port-forward", "pod/" + podName, ports, "-n", namespace, "--kubeconfig", d.kubeconfig}
+	cmdPortFwd := exec.Command(d.cmd, argPortFwd...)
+	err := cmdPortFwd.Start()
+	if err != nil {
+		return nil, err
```
svetasmirnova: You removed error logging. Do you handle or log this error somewhere?

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -561,27 +671,27 @@ func (d *Dumper) getPodSummary(resource, podName, crName string, namespace strin
...
+
 	defer func() {
 		err := cmdPortFwd.Process.Kill()
 		if err != nil {
 			d.logError(err.Error(), "kill port-forward")
 		}
 	}()
 
-	time.Sleep(3 * time.Second) // wait for port-forward command
+	time.Sleep(10 * time.Second) // wait for port-forward command
```
svetasmirnova: Why sleep time increase?

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -497,6 +518,95 @@ func (d *Dumper) getIndividualFiles(namespace string, podName, path, location st
...
+	if err != nil {
+		return "", err
+	}
+	fmt.Printf("DBG: running parseEnvs output: %v\n", string(out))
+
+	resolved := strings.TrimSpace(string(out))
+	return resolved, nil
+}
+
+func (d *Dumper) getAllFilesFormDirectory(namespace string, podName, path, location string, tw *tar.Writer) error {
```
svetasmirnova: Rename function to `getAllFilesFromDirectory`

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -408,9 +426,11 @@ func (d *Dumper) DumpCluster() error {
 
 // runCmd run command (Dumper.cmd) with given args, return it output
 func (d *Dumper) runCmd(args ...string) ([]byte, error) {
+	baseArgs := []string{"--kubeconfig", d.kubeconfig}
+	baseArgs = append(baseArgs, args...)
```
svetasmirnova: If you prefix arguments with --kubeconfig and d.kubeconfig is empty, the following argument will be treated as a path to --kubeconfig. This is wrong.

svetasmirnova (changes requested):

src/go/pt-k8s-debug-collector/dumper/dumper.go
BON4: Left some code after debugging.

src/go/pt-k8s-debug-collector/dumper/dumper.go
BON4: This comes from pod manifest.
```
    Environment:
      MODE:                               postgres
      PGHA_PG_PORT:                       5432
      PGHA_USER:                          postgres
      PATRONI_POSTGRESQL_DATA_DIR:        /pgdata/cluster1-repl1
      PGBACKREST_STANZA:                  db
      PGBACKREST_REPO1_HOST:              cluster1-backrest-shared-repo
      BACKREST_SKIP_CREATE_STANZA:        true
      PGHA_PGBACKREST:                    true
      PGBACKREST_REPO1_PATH:              /backrestrepo/cluster1-backrest-shared-repo
      PGBACKREST_DB_PATH:                 /pgdata/cluster1-repl1
                                          ^
```

src/go/pt-k8s-debug-collector/dumper/dumper.go
BON4: Missed that one.

src/go/pt-k8s-debug-collector/dumper/dumper.go
BON4: It was too little on my machine. Maybe we should implement some tcp checker to confirm that port-forward is working?

src/go/pt-k8s-debug-collector/dumper/dumper.go
BON4: Added an if condition for that. We need this to be prefixed because `exec` command will be wrong.

src/go/pt-k8s-debug-collector/dumper/dumper.go
svetasmirnova: This would be the best.

src/go/pt-k8s-debug-collector/dumper/dumper.go
BON4: Fixed by `pt-k8s-debug-collector`refactor.

src/go/pt-k8s-debug-collector/dumper/dumper.go
BON4: Fixed by `pt-k8s-debug-collector`refactor.

BON4:
Closing due to PR history mess.

PR #1057, opened 2026-01-16
Fix field misalignment in pt-archiver when using --bulk-insert with special characters

xmj147220591:
## Problem

When using `--bulk-insert` option with `pt-archiver`, data containing special characters (tabs, newlines, backslashes) would cause field misalignment during archiving. The issue occurred because the `escape()` function used a single regex replacement `s/([\t\n\\])/\\$1/g` which could process characters in an incorrect order, leading to improper escaping.

For example, if a field contained both a backslash and a tab character, the escaping order could result in incorrect field boundaries being detected by MySQL's `LOAD DATA` statement, causing subsequent fields to be misaligned.

## Solution

Fixed the `escape()` function in `bin/pt-archiver` to escape special characters in the correct sequence:

1. **Escape backslash first** (`\` → `\\`) - This prevents double-escaping issues
2. **Then escape newline** (`\n` → `\\n`) 
3. **Finally escape tab** (`\t` → `\\t`)

The change replaces the single regex replacement with sequential replacements, ensuring that each special character is properly escaped before the next one is processed.

## Changes

- Modified `escape()` function in `bin/pt-archiver` (lines 7519-7523)
- Changed from single regex `s/([\t\n\\])/\\$1/g` to sequential replacements
- Ensures proper escaping order: backslash → newline → tab

## Testing

Tested with data containing various special characters:
- Quoted strings: `'John "Doe"'`
- Tabs and newlines: `'Software Engineer\t\n'`
- Backslashes: `'Backslash\\Test\n'`
- Mixed special characters: `'Quote\'s Test\\t\n'`

All fields now align correctly without misalignment when using `--bulk-insert`.

## Impact

- **Minimal change**: Only modifies the escape logic, no other functionality affected
- **Backward compatible**: Does not change the output format or behavior for normal cases
- **Fixes data corruption**: Prevents field misalignment that could lead to data integrity issues

### DIFF
--- a/bin/pt-archiver
+++ b/bin/pt-archiver
@@ -7516,7 +7516,11 @@ sub escape {
    $optionally_enclosed_by ||= '';
 
    return join($fields_separated_by, map {
-      s/([\t\n\\])/\\$1/g if defined $_;  # Escape tabs etc
+      if (defined $_){
+         s/\\/\\\\/g;
+         s/\n/\\n/g;
+         s/\t/\\t/g;
+      }
       my $s = defined $_ ? $_ : '\N';             # NULL = \N
       # var & ~var will return 0 only for numbers
       if ($s !~ /^[0-9,.E]+$/  && $optionally_enclosed_by eq '"' && $s ne '\N') {
--- a/t/pt-archiver/samples/bulk_insert_special_chars.sql
+++ b/t/pt-archiver/samples/bulk_insert_special_chars.sql
@@ -1,0 +1,22 @@
+-- Schema for testing pt-archiver --bulk-insert with special characters
+-- (tabs, newlines, backslashes) to ensure no field misalignment.
+DROP DATABASE IF EXISTS `bulk_escape`;
+CREATE DATABASE `bulk_escape`;
+
+CREATE TABLE `bulk_escape`.`source` (
+  `id` int unsigned NOT NULL AUTO_INCREMENT,
+  `name` varchar(500) DEFAULT NULL,
+  `job` varchar(450) DEFAULT NULL,
+  `stu_id` int DEFAULT NULL,
+  `title` varchar(450) DEFAULT NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `bulk_escape`.`dest` (
+  `id` int unsigned NOT NULL AUTO_INCREMENT,
+  `name` varchar(500) DEFAULT NULL,
+  `job` varchar(450) DEFAULT NULL,
+  `stu_id` int DEFAULT NULL,
+  `title` varchar(450) DEFAULT NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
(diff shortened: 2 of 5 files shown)


svetasmirnova:
Thank you for the contribution. Please add a test case as described in CONTRIBUTING.md

t/pt-archiver/samples/bulk_insert_special_chars.sql
```diff
@@ -0,0 +1,22 @@
+-- Schema for testing pt-archiver --bulk-insert with special characters
+-- (tabs, newlines, backslashes) to ensure no field misalignment.
+DROP DATABASE IF EXISTS `bulk_escape`;
+CREATE DATABASE `bulk_escape`;
+
+CREATE TABLE `bulk_escape`.`source` (
+  `id` int unsigned NOT NULL AUTO_INCREMENT,
```
svetasmirnova:
```suggestion
  `id` int unsigned NOT NULL,
```
By default, pt-archiver "Adds an extra WHERE clause to prevent pt-archiver from removing the newest row when ascending a single-column AUTO_INCREMENT key. This guards against re-using AUTO_INCREMENT values if the server restarts, and is enabled by default." So either test-file should specify option --nosafe-auto-increment, or source file should not have auto-incremented primary key.

t/pt-archiver/samples/bulk_insert_special_chars.sql
```diff
@@ -0,0 +1,22 @@
...
+  `name` varchar(500) DEFAULT NULL,
+  `job` varchar(450) DEFAULT NULL,
+  `stu_id` int DEFAULT NULL,
+  `title` varchar(450) DEFAULT NULL,
+  PRIMARY KEY (`id`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
+
+CREATE TABLE `bulk_escape`.`dest` (
+  `id` int unsigned NOT NULL AUTO_INCREMENT,
```
svetasmirnova:
```suggestion
  `id` int unsigned NOT NULL,
```
By default, pt-archiver "Adds an extra WHERE clause to prevent pt-archiver from removing the newest row when ascending a single-column AUTO_INCREMENT key. This guards against re-using AUTO_INCREMENT values if the server restarts, and is enabled by default." So either test-file should specify option --nosafe-auto-increment, or source file should not have auto-incremented primary key.

svetasmirnova (changes requested):

svetasmirnova:
Thank you for the adjustments. I provided two suggestions and adjustments for existing tests. Please review and let me know if you are OK with it.

svetasmirnova (approved):

saikumar-vs (approved):

bin/pt-archiver
```diff
@@ -7516,7 +7516,11 @@ sub escape {
    $optionally_enclosed_by ||= '';
 
    return join($fields_separated_by, map {
-      s/([\t\n\\])/\\$1/g if defined $_;  # Escape tabs etc
+      if (defined $_){
```
EvgeniyPatlan, line 7519: maybe we need to try something like this?
my %esc = ("\\" => '\\\\', "\n" => '\n', "\t" => '\t');
return join($fields_separated_by, map {
s/([\\\n\t])/$esc{$1}/g if defined;
........

svetasmirnova: I think we can skip this optimization for 3 lines.

PR #1064, opened 2026-01-28
PT-2368 - full refactor of a config library

BON4:
# Rewrite config library to comply with Percona Toolkit specification

## Summary

Complete rewrite of the config library to fully comply with the [Percona Toolkit configuration file specification](https://www.percona.com/doc/percona-toolkit/LATEST/configuration_files.html). The previous implementation was a simple key-value parser that didn't support many required features.

### New Implementation
- Full Percona Toolkit specification compliance
- Integration with kong CLI parser - configs and command-line args work together
- `--config` flag support - must be first argument, space-separated
- Passthrough arguments - everything after `--` passed to mysql/mysqladmin
- `no-` prefix handling - `no-optimize` sets `optimize=false`
- Proper config file precedence - system - user - tool-specific
- Custom types - `BoolYN` (yes/no/1/0/on/off), `StdinRequestString` (password prompting)
- Error handling
- Full test coverage

- [x] Documentation updated
- [x] Test suite update


src/go/lib/config/old_config.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func (c *Config) GetBool(key string) bool {
+	if val, ok := c.options[key]; ok {
+		if v, ok := val.(bool); ok {
+			return v
+		}
+	}
+	return false
+}
```
JiriCtvrtka, line 71: Could we try improve this with some generic method? I think it should be like:
```
func (c *Config) Get[T any](key string, zero T) T {
    if val, ok := c.options[key]; ok {
        if v, ok := val.(T); ok {
            return v
        }
    }
    return zero
}
```
Usage would be like: 
```
s := cfg.Get[string]("my_string_key", "")
i := cfg.Get[int64]("my_int_key", 0)
f := cfg.Get[float64]("my_float_key", 0)
b := cfg.Get[bool]("my_bool_key", false)
```

src/go/lib/config/old_config.go
```diff
@@ -0,0 +1,147 @@
...
+import (
+	"bufio"
+	"os"
+	"os/user"
+	"strconv"
+	"strings"
+)
+
+type Config struct {
+	options map[string]interface{}
```
JiriCtvrtka, line 34:
```suggestion
	options map[string]any
```

src/go/lib/config/old_config.go
```diff
@@ -0,0 +1,147 @@
...
+}
+
+func DefaultConfig(toolname string) *Config {
+	files, _ := DefaultConfigFiles(toolname)
+	return NewConfig(files...)
+}
+
+func NewConfig(files ...string) *Config {
+	config := &Config{
+		options: make(map[string]interface{}),
```
JiriCtvrtka, line 106:
```suggestion
		options: make(map[string]any),
```

src/go/lib/config/old_config.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+	for _, filename := range files {
+		if _, err := os.Stat(filename); err == nil {
+			read(filename, config.options)
+		}
+	}
+	return config
+}
+
+func read(filename string, opts map[string]interface{}) error {
```
JiriCtvrtka, line 116:
```suggestion
func read(filename string, opts map[string]any) error {
```

src/go/lib/config/config.go
```diff
@@ -2,146 +2,338 @@ package config
...
+
+	"github.com/alecthomas/kong"
 )
 
-type Config struct {
-	options map[string]interface{}
-}
+// BoolYN represents a boolean flag that accepts multiple textual representations.
+// Supported true values: 1, true, yes, y, on, "" (empty)
+// Supported false values: 0, false, no, n, off
```
JiriCtvrtka, line 33: TBH, I’m not a fan of this. It also increases the complexity of the rest of the code. But if this is the logic behind the rest of Percona Toolkit, let’s implement it to keep it aligned with the rest of the project.

svetasmirnova:
Options in the only modified tool do not work as expected:

```
sveta@s76:~/src/percona/percona-toolkit$ ./bin/pt-k8s-debug-collector --version
pt-k8s-debug-collector: error: couldn't find binding of type []interface {} for parameter 0 of func(...interface {}) error(), use kong.Bind([]interface {})
sveta@s76:~/src/percona/percona-toolkit$ ./bin/pt-k8s-debug-collector version
pt-k8s-debug-collector: error: unexpected argument version
```

Expected behavior:

```
$ ./bin/pt-mongodb-summary --version
pt-mongodb-summary
Version 3.7.1
Build: 2026-01-30T15:30:33+0300 using go1.25.5
Commit: 14324ff374b078089e3bbae2bdcf6e09f7c4cc35
```

And we should not change this format, because it is aligned with other tools.

svetasmirnova (changes requested):
See comment above.

src/go/lib/config/old_config.go
BON4: I thought we would disregard `old_config.go`, when all pt-tools will be migrated to a new config lib.

BON4:
@svetasmirnova Yep, forget to push some fixes.

BON4:
@JiriCtvrtka Not sure if we should change `old_config.go` because this is old implementation and eventually it will be replaced with the new one. I left it just to support compatibility with not yet migrated tools.

src/go/lib/config/config.go
BON4: I've implemented the `BoolYN` type to maintain backward compatibility with our existing config.go library. It handled it specifically for config files: https://github.com/percona/percona-toolkit/blob/7683c460c44e2386f1ec7c3823d6b2704b61e661/src/go/lib/config/old_config.go#L119-L126

@svetasmirnova Should we keep supporting this?

svetasmirnova: Let's keep it for compatibility with Perl tools and less user surprise. Also, if we decide to migrate some of them into Go, this functionality will be required.

src/go/lib/config/old_config.go
```diff
@@ -0,0 +1,147 @@
+package config
```
svetasmirnova, line 23: Please add a header here, so this file is left only for backward compatibility, deprecated, and read-only.

src/go/lib/config/config.go
```diff
@@ -2,146 +2,338 @@ package config
...
-		}
-		if lcval == "false" || lcval == "no" {
-			opts[key] = false
+	return &configFile{
+		options:     &optsBuffer,
+		passthrough: passthrough,
+	}, nil
+}
+
+// Setup initializes kong parser with Percona Toolkit configuration file support.
```
svetasmirnova, line 213: We need to document this flag in README.rst for each tool

svetasmirnova (changes requested):
Please also merge changes from 3.x, so library files have a copyright header, and add a copyright header to the new config.go

src/go/pt-galera-log-explainer/main.go
```diff
@@ -39,29 +39,55 @@ var CLI struct {
...
 
+func (c *CliOptions) AfterApply() error {
+	if c.VersionCheck {
+		advice, err := versioncheck.CheckUpdates(toolname, Version)
+		if err != nil {
+			log.Error().Msgf("cannot check version updates: %s", err.Error())
+		} else if advice != "" {
+			log.Info().Msgf("%s", advice)
+		}
+	}
```
ylacancellera, line 78: This is currently breaking tests with
```
        + 	time="2026-02-16T18:19:16+01:00" level=info msg="Checking for updates"
        + 	time="2026-02-16T18:19:16+01:00" level=info msg="Contacting version check API at https://v.percona.com/. Timeout set to 3s"
```
Is this really testing the API call during tests ? I can update the expected outputs with those once we're ready.

src/go/lib/config/old_config.go
svetasmirnova: I understand that this kept only for backward compatibility and planned to be removed. No need to fix code here. @BON4  please add a comment that this will be deprecated and removed in future versions.

src/go/lib/config/old_config.go
svetasmirnova: This is old config planned to be removed, no need to fix

src/go/lib/config/old_config.go
svetasmirnova: This is old config planned to be removed, no need to fix

src/go/lib/config/old_config.go
svetasmirnova: This is old config planned to be removed, no need to fix

src/go/pt-galera-log-explainer/README.rst
```diff
@@ -99,6 +99,9 @@ Will print every implemented regexes:
 Available flags
 ~~~~~~~~~~~~~~~
 
+``--config``
+    List of Percona Toolkit configuration file(s) separeted by comma without equal sign. Must be a first flag. Uses default config file locations if not specified.
```
svetasmirnova:
```suggestion
    List of Percona Toolkit configuration file(s) separated by a comma without an equal sign. Must be a first flag. Uses default config file locations if not specified.
```

src/go/pt-k8s-debug-collector/README.rst
```diff
@@ -125,6 +125,10 @@ Usage
 Supported Flags
 ================
 
+``--config``
+
+List of Percona Toolkit configuration file(s) separeted by comma without equal sign. Must be a first flag. Uses default config file locations if not specified.
```
svetasmirnova:
```suggestion
List of Percona Toolkit configuration file(s) separated by a comma without an equal sign. Must be a first flag. Uses default config file locations if not specified.
```

src/go/pt-galera-log-explainer/main.go
svetasmirnova: @ylacancellera, This error occurs because the VersionCheck integration is incomplete across all tools. We will fix it later before the 3.7.2 release. Please ignore this for now; later, we will create a solution to skip the version check for tests when it is not needed.

svetasmirnova (changes requested):
Please answer questions and process suggested changes.

src/go/pt-k8s-debug-collector/README.rst
svetasmirnova: Typo still exists

svetasmirnova (changes requested):

ylacancellera (approved):

svetasmirnova (approved):

PR #1071, opened 2026-02-13
PT-2352 - make pt-mongodb-summary accept standard connection syntax

BON4:
This commit adds `--uri` flag so it can be used as connection string for mongodb.

- [x] Documentation updated
- [x] Test suite update

### DIFF
--- a/src/go/pt-mongodb-summary/README.rst
+++ b/src/go/pt-mongodb-summary/README.rst
@@ -32,14 +32,32 @@ For better results, host must be a **mongos** server.
 Options
 -------
 
-``-a``, ``--auth-db``
+``-a``, ``--authenticationDatabase``
   Specifies the database used to establish credentials and privileges
   with a MongoDB server.
   By default, the ``admin`` database is used.
 
+``-c``, ``--no-version-check``
+  Disables checking the version of MongoDB before running the report.
+
 ``-f``, ``--output-format``
   Specifies the report output format. Valid options are: ``text``, ``json``.
   The default value is ``text``.
+
+``-h``, ``--help``
+  Show help message and exit.
+
+``--host``
+  Specifies the hostname or IP address of the MongoDB server to connect to.
+
+``-i``, ``--running-ops-interval``
+  Interval in milliseconds to wait between samples of running operations.
+  Default: 1000 milliseconds.
+
+``-l``, ``--log-level``
+  Specifies the logging level. Valid options: ``panic``, ``fatal``, ``error``, 
+  ``warn``, ``info``, ``debug``.
+  Default: ``error``.
 
 ``-p``, ``--password``
   Specifies the password to use when connecting to a server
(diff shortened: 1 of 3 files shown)


src/go/pt-mongodb-summary/main.go
```diff
@@ -170,6 +172,7 @@ type cliOptions struct {
 	SSLPEMKeyFile      string
 	RunningOpsSamples  int
 	RunningOpsInterval int
+	URI                string
 	Help               bool
```
svetasmirnova, line 176: We need to document new options in the README.rst

svetasmirnova (changes requested):

svetasmirnova (approved):

src/go/pt-mongodb-summary/README.rst
```diff
@@ -32,28 +32,67 @@ For better results, host must be a **mongos** server.
...
+``--sslCAFile``
+  Path to the SSL CA certificate file used for authentication.
+
+``--sslPEMKeyFile``
+  Path to the SSL client PEM file used for authentication.
+
+``--uri``
+  Full MongoDB URI describing hosts and options.
+  Command-line flags have higher priority than URI settings.
+  If a full URI is provided, you cannot also specify ``--host`` or ``--port``.
```
igroene, line 87: what about username and pass arguments? does the URI support having them? does it override them??

igroene (changes requested):
let's document a bit better the URI format. Provide a quick example:
mongodb://user:pass@mongodb:port/

src/go/pt-mongodb-summary/main.go
```diff
@@ -933,6 +933,10 @@ func parseFlags() (*cliOptions, error) {
 	gop.BoolVarLong(&opts.Version, "version", 'v', "", "Show version & exit")
 	gop.BoolVarLong(&opts.NoVersionCheck, "no-version-check", 'c', "", "Default: Don't check for updates")
 
+	gop.StringVarLong(&opts.URI, "uri", 0, "URI describes the hosts to be used and options. Flags has higher priority.")
```
artemgavrilov: > Flags has higher priority

That means that dedicated `host` and `port` flags will win?

src/go/pt-mongodb-summary/main.go
```diff
@@ -958,7 +962,14 @@ func parseFlags() (*cliOptions, error) {
 	gop.Parse(os.Args)
 
 	if gop.NArgs() > 0 {
-		opts.Host = gop.Arg(0)
+		if gop.IsSet("host") || gop.IsSet("port") {
```
artemgavrilov: Shouldn't we check here also for `--uri` flag?

BON4: Yes, absolutely. Good catch. 
Fixed.

src/go/pt-mongodb-summary/main.go
BON4: I've extended help message via --help command, to make it clear that `host` and `port` is incompatible with `uri`.

src/go/pt-mongodb-summary/README.rst
BON4: All flags, except `--host` and `--port`. If username set in `--uri` and `--username` flag is used, flag will override the username in `--uri`. This is the way it is done in `mongosh`, I think it will be familiar to the mongo team.

src/go/pt-mongodb-summary/main.go
BON4: I think this is fixed.

igroene (approved):

artemgavrilov (approved):

PR #1079, opened 2026-02-19
Added users data to history table

losgrandes:
### Background
Idea behind this PR is to add information about DB users who executed a query to the history table. I've decided to add JSON column to store array of users for each query. There are probably more elegant solutions, but I wanted to keep "one table" concept.

### Summary of changes
- added optional usernames JSON type column with support for it in history table
- updated documentation
- updated existing and added new tests case

### DIFF
--- a/bin/pt-query-digest
+++ b/bin/pt-query-digest
@@ -9405,6 +9405,11 @@ has history_metrics => (
    isa => 'ArrayRef',
 );
 
+has main_columns => (
+   is => 'rw',
+   isa => 'ArrayRef',
+);
+
 has column_pattern => (
    is       => 'ro',
    isa      => 'Regexp',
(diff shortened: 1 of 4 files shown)


svetasmirnova:
Thank you for the contribution. You modified auto-generated code that will be rewritten after we run , ./utils/update-modules. Please add your changes into lib/QueryHistory.pm, then update modules in all tools as described in CONTRIBUTING.md

svetasmirnova (changes requested):
Please update lib/QueryHistory.pm

losgrandes:
> Thank you for the contribution. You modified auto-generated code that will be rewritten after we run , ./utils/update-modules. Please add your changes into lib/QueryHistory.pm, then update modules in all tools as described in CONTRIBUTING.md

My apologies for overlooking this.
I've run `for t in bin/*; do ./util/update-modules ${t} QueryHistory; done` and committed these changes. Should be good.
Thanks.

svetasmirnova:
Changes look good for me, thank you!

But MySQL user accounts consist of two parts: usernames and hosts. E.g., sveta@percona.com is not the same user as sveta@my.blog. Maybe it makes sense to store host part too?

bin/pt-query-digest
svetasmirnova:
```suggestion
  CREATE TABLE query_history (
```

svetasmirnova (changes requested):
Please fix this stalled table name in the user manual part of the code too (query_review_history -> query_history).

losgrandes:
> Changes look good for me, thank you!
> 
> But MySQL user accounts consist of two parts: usernames and hosts. E.g., [sveta@percona.com](mailto:sveta@percona.com) is not the same user as [sveta@my.blog](mailto:sveta@my.blog). Maybe it makes sense to store host part too?

This is very good idea, Sveta. In this code change I have reused already available user data which doesn't contain host information.
Adding that info requires reworking tools' core logic, which I would like to avoid currently.
I think it would be safer to add it later. WDYT?

svetasmirnova (changes requested):
Host is defined as part of `$ea->results->{'classes'}{$item}`, so I do not think it will require rewriting the tool's core logic. I suggest adding it now, together with the new "user" column.

Please also fix formatting: Percona Toolkit uses 3 spaces for indentation.

PR #1091, opened 2026-03-21
Swap to term.ReadPassword to remove outdated gopass dependency and stop using ssh/terminal

lvoytek:
The [howeyc/gopass](https://github.com/howeyc/gopass) dependency has been archived and unmaintained for around 5 years now, and recommends using `golang.org/x/crypto/ssh/terminal` as a replacement. This PR replaces `gopass.GetPasswd()` instances with `terminal.ReadPassword(0)` to allow the dependency to be removed. This is what is currently done in `pt-secure-collect`, making this a bit more consistent across tools too. 

- [ ] Documentation updated
- [ ] Test suite update

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,6 @@ require (
 	github.com/google/go-cmp v0.7.0
 	github.com/google/uuid v1.6.0
 	github.com/hashicorp/go-version v1.8.0
-	github.com/howeyc/gopass v0.0.0-20210920133722-c8aef6fb66ef
 	github.com/lib/pq v1.12.0
 	github.com/mattn/go-shellwords v1.0.12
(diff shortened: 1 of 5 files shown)


svetasmirnova (changes requested):
Terminal is also deprecated:

> Deprecated: this package moved to golang.org/x/term. 

Likely need to change to https://pkg.go.dev/golang.org/x/term#ReadPassword instead.

lvoytek:
That's fair, updated all of the password entry locations for consistency

BON4 (approved):
Hi! Thanks for the PR. Great catch on the archived dependency. It's always good to trim down the technical debt.

svetasmirnova (approved):

PR #1054, opened 2026-01-09
PT-2488 - full refactor of a pt-k8s-debug-collector

BON4:
This refactor includes replacing all of the `kubectl` cli calls with golang sdk for k8s. Additionaly dumper now has new structure, new logger, tar file path controll, and multithreaded approach for downloading and exporting files form multiple pods.

### Resulting Archive changes
- added `cluster-scoped` folder for the resources that are cluster wide. 
- in the root there is a total log file, instead of just errors.

- [x] The contributed code is licensed under GPL v2.0
- [x] Contributor Licence Agreement (CLA) is signed
- [x] util/update-modules has been ran
(`/lib` was not changed)
- [x] Documentation updated
- [x] Test suite update

PR #1101, opened 2026-03-31
PT-2218 - Add test to check per container logs for pxc

BON4:
This issue have been fixed since #1054 was merged. This PR introduces modification of already existing test, to check if logs are indeed separated by container for `pxc`.
- [x] Documentation updated
- [x] Test suite update

### DIFF
--- a/src/go/pt-k8s-debug-collector/main_test.go
+++ b/src/go/pt-k8s-debug-collector/main_test.go
@@ -328,6 +328,43 @@ func (s *CollectorSuite) TestIndividualFiles() {
 				return in[:nl]
 			},
 		},
+		{
+			namespace: "pxc",
+			// If pod logs are exported as one file per container
+			name: "pxc_container_logs_split_by_container",
+			// tar -tf cluster-dump.tar.gz --wildcards 'cluster-dump/pxc/*/*.log'
+			cmd:  []string{"tar", "-tf", "cluster-dump.tar.gz", "--wildcards", "cluster-dump/pxc/*/*.log"},
+			want: []string{"logrotate.log", "logs.log", "pxc-init.log", "pxc.log"},
+			preprocessor: func(in string) string {
+				required := map[string]bool{
+					"logrotate.log": true,
+					"logs.log":      true,
+					"pxc-init.log":  true,
+					"pxc.log":       true,
+				}
+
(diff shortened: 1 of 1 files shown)


svetasmirnova (approved):

src/go/pt-k8s-debug-collector/main_test.go
```diff
@@ -328,6 +328,43 @@ func (s *CollectorSuite) TestIndividualFiles() {
...
+					if len(parts) != 2 {
+						continue
+					}
+
+					b := parts[1]
+					if !required[b] {
+						continue
+					}
+
+					if !slices.Contains(result, b) && b != "." && b != "" {
```
artemgavrilov: Not sure under what circumstances `b` may be dot or empty here if we explicitly compare it with `required` few lines above?

src/go/pt-k8s-debug-collector/main_test.go
```diff
@@ -328,6 +328,43 @@ func (s *CollectorSuite) TestIndividualFiles() {
...
+			// tar -tf cluster-dump.tar.gz --wildcards 'cluster-dump/pxc/*/*.log'
+			cmd:  []string{"tar", "-tf", "cluster-dump.tar.gz", "--wildcards", "cluster-dump/pxc/*/*.log"},
+			want: []string{"logrotate.log", "logs.log", "pxc-init.log", "pxc.log"},
+			preprocessor: func(in string) string {
+				required := map[string]bool{
+					"logrotate.log": true,
+					"logs.log":      true,
+					"pxc-init.log":  true,
+					"pxc.log":       true,
+				}
```
artemgavrilov, line 348: Minor: It's better to implement sets like `map[string]struct{}` and use _, ok := required[b] to check if value is there.

artemgavrilov (approved):

src/go/pt-k8s-debug-collector/main_test.go
BON4: Agreed, it is a excess defensive check. I will remove it.

src/go/pt-k8s-debug-collector/main_test.go
BON4: Done.

PR #1103, opened 2026-04-02
PT-2189 - Fix cluster selection

BON4:
Fix `--cluster` handling for `pt-k8s-debug-collector`

This change makes `--cluster` select the target kubeconfig context when building the Kubernetes clients. Existing auto-detection behavior is preserved, but it now runs against the chosen cluster. If `--cluster` is not provided, the current context is still used. For backward compatibility, parsing of `resource/cluster` style values is retained internally.

Also adds unit coverage for resource-spec parsing and context-based REST config selection.

- [ ] Documentation updated
- [x] Test suite update


svetasmirnova (approved):

src/go/pt-k8s-debug-collector/dumper/dumper.go
```diff
@@ -84,14 +84,17 @@ type exportJob struct {
...
+func New(location, namespace, kubeconfig, clusterName, forwardport, resource string, skipPodSummary bool, concurrentExportWorkers int) (*Dumper, error) {
 	safeLog := NewSafeLogger()
 
 	log.AddHook(&ErrorArchiveHook{safeLogger: safeLog})
 
-	config, err := clientcmd.BuildConfigFromFlags("", kubeconfig)
+	_, resourceClusterName := parseResourceSpec(resource)
+	if clusterName == "" {
+		clusterName = resourceClusterName
+	}
```
artemgavrilov: It seems it can be simplified like this:
```suggestion
	if clusterName == "" {
		_,  clusterName = parseResourceSpec(resource)
	}
```

artemgavrilov (approved):

PR #1110, opened 2026-04-09
PT-2448 - Redact pgbouncer secret references in pt-k8s-debug-collector

eslavyansky:
Redact pgbouncer secret references in pt-k8s-debug-collector                                                                                                                                                            
                                                                                                                                                                                                                                   
pt-k8s-debug-collector now strips volumes and volumeMounts entries that reference pgbouncer secrets from exported Kubernetes resources, preventing sensitive pgbouncer data from leaking into the diagnostic archive.            
                                                                                                                                                                                                                                   
Changes:                                                                                                                                                                                                                         

- dumper/dumper.go - added redactPgbouncerVolumeRefs, redactPodSpec, and hasPgbouncerSecretRef functions that, during generic resource export, remove volume entries referencing pgbouncer secrets (via secret.secretName or     projected.sources) from pod and pod template specs, along with the corresponding volumeMounts across all container types.                                                                                                        
- main_test.go - added integration test TestPgBouncerSecretsNotCollected that verifies the output archive contains no pgbouncer-frontend.ca-roots entries for pgo and pgv2 namespaces.

- [ ] Documentation updated
- [x] Test suite update

### DIFF
--- a/src/go/pt-k8s-debug-collector/main_test.go
+++ b/src/go/pt-k8s-debug-collector/main_test.go
@@ -707,9 +707,7 @@ func (s *CollectorSuite) TestRequiredFilesExist() {
 	}
 }
 
-/*
-PT-2448 - pt-k8s-debug-collector should not collect secret details of pgbouncer
-*/
+// PT-2448 - pt-k8s-debug-collector should not collect secret details of pgbouncer
 func (s *CollectorSuite) TestPgBouncerSecretsNotCollected() {
 	if s.Namespace != "pgo" && s.Namespace != "pgv2" {
 		s.T().Skip("Only applicable to pgo and pgv2 namespaces")
@@ -722,9 +720,10 @@ func (s *CollectorSuite) TestPgBouncerSecretsNotCollected() {
 				"--forwardport", s.ForwardPort,
 				"--resource", resource,
 				"--skip-pod-summary")
-			_ = cmd.Run()
+			err := cmd.Run()
+			s.NoError(err)
 
-			testcmd := "tar -xf cluster-dump.tar.gz --to-command 'grep \"pgbouncer-frontend.ca-roots\"' 2>/dev/null | wc -l"
+			testcmd := "tar -xf cluster-dump.tar.gz --to-command 'grep \"pgbouncer-frontend\"' 2>/dev/null | wc -l"
 			out, err := exec.Command("sh", "-c", testcmd).Output()
 			s.NoError(err)
 			s.Equal("0", strings.TrimSpace(string(out)), "Should not find pgbouncer secret details in archive files")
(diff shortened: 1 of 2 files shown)


svetasmirnova:
So now this file cluster1-pgbouncer.yaml not collected at all? Let me check with the bug reporter if this was the desired behavior.

svetasmirnova (approved):

svetasmirnova:
Reporter confirmed this is expected behavior.

src/go/pt-k8s-debug-collector/main_test.go
```diff
@@ -706,3 +706,28 @@ func (s *CollectorSuite) TestRequiredFilesExist() {
...
+	}
+
+	for _, resource := range s.Resources {
+		s.Run("Resource_"+resource, func() {
+			cmd := exec.Command(TOOL_PATH,
+				"--kubeconfig", s.KubeConfig,
+				"--forwardport", s.ForwardPort,
+				"--resource", resource,
+				"--skip-pod-summary")
+			_ = cmd.Run()
```
artemgavrilov: Better to always check for an error

src/go/pt-k8s-debug-collector/main_test.go
```diff
@@ -706,3 +706,28 @@ func (s *CollectorSuite) TestRequiredFilesExist() {
...
+
+			testcmd := "tar -xf cluster-dump.tar.gz --to-command 'grep \"pgbouncer-frontend.ca-roots\"' 2>/dev/null | wc -l"
```
artemgavrilov: Why are you grep here exactly for this file and not everything that has `pgbouncer` in the name?

src/go/pt-k8s-debug-collector/main_test.go
```diff
@@ -706,3 +706,28 @@ func (s *CollectorSuite) TestRequiredFilesExist() {
 		}
 	}
 }
+
+/*
+PT-2448 - pt-k8s-debug-collector should not collect secret details of pgbouncer
+*/
```
artemgavrilov: Minor: It seems that rest of the code here uses `//` style of comments

src/go/pt-k8s-debug-collector/main_test.go
eslavyansky: Narrowed to pgbouncer-frontend on purpose the archive legitimately keeps the `pgbouncer` Service/Deployment/labels/config/CR, which we want collected. Only `pgbouncer-frontend.{ca-roots,crt,key}` is sensitive and gets redacted, so a plain `pgbouncer` grep would match the legitimate resources and fail. The pgbouncer-frontend prefix now covers all three leaked keys, not just .ca-roots. Verified on a live pgv2 cluster.

artemgavrilov (approved):

PR #1111, opened 2026-04-09
PT-2500 Fix TestVersionOption regex to accept release suffix in version

eslavyansky:
Updated the TestVersionOption regex in all tools to accept version strings with an optional pre-release suffix (e.g. 3.6.0-1). The old pattern v?\d+.\d+.\d+ did not match versions which appends a - suffix.

Tests for all tools are affected - done

- [X] Documentation updated
- [X] Test suite update


svetasmirnova (approved):

artemgavrilov:
@eslavyansky Usually pre-release versions may also contain characters, like `3.6.0-rc1`, what about that scenario? Also if we expect only semver here, maybe it's worth to use official regular expression? 
https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

svetasmirnova:
@artemgavrilov, they are not pre-release suffixes; this is how we release security patches based on earlier released versions.

@eslavyansky, please still evaluate the regular expression suggestion.

artemgavrilov (approved):

PR #1118, opened 2026-04-18
PT-2272 - Fix too many open files error

thealexjo:
Fix `Too many open files` error that often occurs due to not-closed replica watching connections.
This also fixes possible connection storm on replicas as it closes unnecessary connections.

- [ ] Documentation updated
- [ ] Test suite update

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -4503,7 +4503,7 @@ sub _find_replicas_by_hosts {
       $sql = 'SHOW SLAVE HOSTS';
       $source_name='master';
    }
-   
+
    PTDEBUG && _d($dbh, $sql);
    @replicas = @{$dbh->selectall_arrayref($sql, { Slice => {} })};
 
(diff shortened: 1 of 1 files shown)


svetasmirnova (changes requested):
You fixed auto-generated code. Changes should go into lib/ReplicaLagWaiter.pm and lib/MasterSlave.pm Otherwise, they will be overridden when we run the update-modules utility next time. Also, a regression test case is required.

See CONTRIBUTING.md for details.

svetasmirnova:
Your fix breaks tests with the error:

```
# Error while waiting for replica lag: DBD::mysql::db selectrow_array failed: Statement not active [for Statement "SELECT @@SERVER_ID"] at /home/sveta/src/percona/percona-toolkit/bin/pt-online-schema-change line 4402.
```

So this is not so simple as ensuring that all connections are closed. Therefore, I will reject this PR and either work with idea presented in https://github.com/percona/percona-toolkit/pull/1123 (has same issue as this PR), or introduce new solution.

PR #1123, opened 2026-05-08
PT-2534 - fixed connection leak due to confusion on parent flag

marcelohpf:
Introduces a new argument to Cxn to flag that connection should not be closed automatically with InactiveDestroy, while keep the `parent` argument for its current usage of topology parent (or connection source parent)

The code had two different purposes to the same argument "parent" - topology parent and process parent in fork scenarios leading to a leak when get_replicas was called multiple times (e.g. when waiting for replication delays).

None of the current percona toolkit bin/ scripts seems to use forking other than deadlock detection. None of the "parent => X" cases were properly used in a multiprocess scenario, even when daemonizing the tool, connections were created after the fork;

You can reproduce it by:
* create large enough table;
* start dev environment mysqls
* trigger OSC:
```
./bin/pt-online-schema-change \
  'D=mysqlslap,t=t1' --alter "ADD COLUMN mycolumn varchar(255) NOT NULL DEFAULT ''" \
  --execute --no-version-check --statistics --no-check-alter --recurse 2 --max-lag 10 \
  --chunk-size 10 --progress time,30  \
  --recursion-method dsn=D=percona,t=dsns \ # important
  --no-check-replication-filters \
  --database mysqlslap --host 127.0.0.1 \
  --user msandbox --password msandbox --port 12345
```

Then connect to one of the read-replicas and watch process list grow indefinitely:
```
while true; do 
  mysql -h 127.0.0.1 -P 12346 -u msandbox -pmsandbox -e 'show processlist;'
  sleep 1;
done
```

The root cause of the issue is that InactiveDestroy prevent dbh closing connection as part of the DESTROY workflow that is triggered when object looses reference count.

```
# other code
my $dbh;
{
  $dbh = DBI::connect();
   $dbh -> {InactiveDestroy} = 1;
  fork();
  if (is children) {
     $dbh->ping();  # ok
     exit;
  }
}
$dbh->ping() # should not fail because InactiveDestroy = 1; else it would
```

- [X] util/update-modules has been ran
- [X] Documentation updated
- [X] Test suite update

### DIFF
--- a/t/lib/Cxn.t
+++ b/t/lib/Cxn.t
@@ -305,7 +305,7 @@ my $pid;
 # Let the child know that we (the parent) have left that ^ code block,
 # so our copy of $parent_cxn has been destroyed, but hopefully the child's
 # copy is still alive, i.e. has an active/not-disconnected dbh.
-diag(`touch $sync_file`);
+PerconaTest::touch_file($sync_file);
 
 # Wait for the child.
 waitpid($pid, 0);
@@ -315,7 +315,7 @@ ok(
    "Child created outfile"
 );
 
-my $output = `cat $outfile 2>/dev/null`;
+my $output = PerconaTest::slurp_file($outfile);
 
 is(
    $output,
@@ -346,13 +346,15 @@ my $children_ping_file = "/tmp/pt-cxn-children-ping.$PID";
       # I am the child.
       # Wait for the parent to leave this code block which will cause
       # the $parent_cxn to be destroyed.
-      open(my $fh, '>', $children_ping_file);
+      open(my $fh, '>', $children_ping_file) or die "Cannot open $children_ping_file: $!";
       print($fh "Children ping: " . ($parent_cxn->dbh->ping() ? "OK": "FAIL") . "\n");
-      diag(`touch $children_sync_file`);
+
+      PerconaTest::touch_file($children_sync_file);
+
       # we wait parent exit this code-block so DESTROY actually disconnects the session
       PerconaTest::wait_for_files($sync_file);
-      print($fh "Children ping after parent exited: " . ($parent_cxn->dbh->ping() ? "OK": "FAIL"));
-      close($fh);
+      print($fh "Children ping after parent exited: " . ($parent_cxn->dbh->ping() ? "OK": "FAIL") . "\n");
+      close($fh) or die "Cannot close $children_ping_file: $!";
       exit;
    }
    PerconaTest::wait_for_files($children_sync_file); # wait confirmation that children conn was ok
@@ -361,16 +363,16 @@ my $children_ping_file = "/tmp/pt-cxn-children-ping.$PID";
 # Let the child know that we (the parent) have left that ^ code block,
 # so our copy of $parent_cxn has been destroyed, but hopefully the child's
 # copy is still alive, i.e. has an active/not-disconnected dbh.
-diag(`touch $sync_file`);
+PerconaTest::touch_file($sync_file);
 
 # Wait for the child.
 waitpid($pid, 0);
 
-my $children_output = `cat $children_ping_file 2>/dev/null`;
+my $children_output = PerconaTest::slurp_file($children_ping_file);
 
 is(
    $children_output,
-   "Children ping: OK\nChildren ping after parent exited: FAIL",
+   "Children ping: OK\nChildren ping after parent exited: FAIL\n",
    'Expected children to succeed on first ping but fail when parent exited'
 );
 
--- a/t/lib/Cxn.t
+++ b/t/lib/Cxn.t
@@ -63,7 +63,7 @@ sub test_var_val {
    my ($dbh, $var, $val, %args) = @_;
 
    my @row;
-   if ( !$args{user_var} ) { 
+   if ( !$args{user_var} ) {
       my $sql = "SHOW " . ($args{global} ? "GLOBAL" : "SESSION " )
               . "VARIABLES LIKE '$var'";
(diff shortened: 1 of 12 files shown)


t/lib/Cxn.t
```diff
@@ -326,6 +326,60 @@ is(
...
+      parent => 'h=127.1,P=12345,u=msandbox,p=msandbox' # fake parent id
+   );
+   $parent_cxn->connect();
+   $pid = fork();
+
+   if ( defined($pid) && $pid == 0 ) {
+      # I am the child.
+      # Wait for the parent to leave this code block which will cause
+      # the $parent_cxn to be destroyed.
+      open(my $fh, '>', $children_ping_file);
```
EvgeniyPatlan:
```suggestion
      open(my $fh, '>', $children_ping_file) or die "Cannot open $children_ping_file: $!";
```

t/lib/Cxn.t
```diff
@@ -326,6 +326,60 @@ is(
...
+      print($fh "Children ping: " . ($parent_cxn->dbh->ping() ? "OK": "FAIL") . "\n");
+      diag(`touch $children_sync_file`);
```
EvgeniyPatlan: Possible injection-prone if the filename ever contains shell metacharacters. Replace with native Perl:open(my t, '>', $children_sync_file) and close(t), or utime undef, undef, $children_sync_file` if the file exists.

t/lib/Cxn.t
```diff
@@ -326,6 +326,60 @@ is(
...
+      # we wait parent exit this code-block so DESTROY actually disconnects the session
+      PerconaTest::wait_for_files($sync_file);
+      print($fh "Children ping after parent exited: " . ($parent_cxn->dbh->ping() ? "OK": "FAIL"));
```
EvgeniyPatlan: Missing trailing \n (the first print has one), so the file ends without a newline

t/lib/Cxn.t
```diff
@@ -326,6 +326,60 @@ is(
...
+      close($fh);
```
EvgeniyPatlan: close($fh) or die "close: $!"

t/lib/Cxn.t
```diff
@@ -326,6 +326,60 @@ is(
...
+      exit;
+   }
+   PerconaTest::wait_for_files($children_sync_file); # wait confirmation that children conn was ok
+   ok($parent_cxn->dbh->ping(), 'Parent connection was fine');
+}
+# Let the child know that we (the parent) have left that ^ code block,
+# so our copy of $parent_cxn has been destroyed, but hopefully the child's
+# copy is still alive, i.e. has an active/not-disconnected dbh.
+diag(`touch $sync_file`);
```
EvgeniyPatlan: Possible injection-prone

EvgeniyPatlan (commented):
I would suggest some changes. They are for test scenario but still it is netter to follow perl recommendations

t/lib/Cxn.t
marcelohpf: Thanks for the suggestion, I added it to other places as well.

t/lib/Cxn.t
marcelohpf: Indeed, I followed other references on the file to follow the "project standards".

I've replaced it by a new function in PerconaTest::touch_file and updated other `cat ` and `touch ` cases as well.

t/lib/Cxn.t
marcelohpf: Added training \n. It is really easy to miss it in perl string formating.

t/lib/Cxn.t
marcelohpf: Added

t/lib/Cxn.t
marcelohpf: Addressed in the other comment, thanks

EvgeniyPatlan (approved):
LGTM

marcelohpf:
I had to update the commit author due to the CLA agreement, sorry for the force push;

EvgeniyPatlan:
thanks you @marcelohpf for the contribution and for the quick fixes!

charles-001:
I'm pretty sure this is the issue I'm currently running into with the latest version (3.7.1). While pt-osc is running, it's creating a ton of connections on the replica to check its lag (like 1000+). Eventually, it hits `max_user_connections` and fails. Seems like a regression happened in 3.7.1

```Cannot connect to MySQL: DBI connect(';mysql_read_default_file=~/.my.cnf;host=host;mysql_read_default_group=client','',...) failed: User my_user already has more than 'max_user_connections' active connections at /usr/bin/pt-online-schema-change line 2351.```

**Comment by svetasmirnova** [comment]
Many tests are failing; I need to investigate more on this PR.

**Comment by marcelohpf** [comment]
@svetasmirnova I've rebased with latest from 3.x. I have some time to help with tests

**Comment by svetasmirnova** [comment]
Tests fail with error:

```
# Error while waiting for replica lag: DBD::mysql::db selectrow_array failed: Statement not active [for Statement "SELECT @@SERVER_ID"] at /home/sveta/src/percona/percona-toolkit/bin/pt-online-schema-change line 4402.
```

This means that the connection closed when it is still needed.

svetasmirnova:
After investigation, I found that the fix for PT-2250 caused this.  Your ideas about InactiveDestroy and multi-thread applications are valid and will be considered in the future. But I decided to fix this issue differently. You can refer to the fix at https://github.com/percona/percona-toolkit/pull/1171. This is a draft PR at the moment because I still need to write an automated test and apply changes in libraries to other tools.

PR #1126, opened 2026-05-12
Pt mongodb stalk

palazuelosagustin:
- [x] Documentation updated
- [x] Test suite update


svetasmirnova (changes requested):
Thank you for the contribution. Please follow the rules:

- Documentation for Bash tools should be placed inline. Check other tools like pt-mysql-summary or pt-stalk. There should not be a separate README file for Bash tools.
- Sign contributor agreement
- Create a regression test for your changes, see examples in the t/TOOLNAME directories

Please also consider if this change fits existing pt-stalk, as a new option.

palazuelosagustin:
Changes made:

  - Removed bin/pt-mongodb-stalk.README.
  - Moved the useful README content into inline POD inside bin/pt-mongodb-stalk.
  - Added regression test: t/pt-mongodb-stalk/pt-mongodb-stalk.t.
  - Test uses fake MongoDB/OS commands, so it does not require a real MongoDB instance.
  - Test covers --no-stalk, expected output files, currentOp allUsers:true, and retained non-empty mongotop.err.
  
  On the pt-stalk question: I’d argue this is better as a separate tool for now. pt-stalk is tightly MySQL-specific around mysql, mysqladmin, SHOW GLOBAL STATUS, SHOW PROCESSLIST, and MySQL option passthrough.
  Folding MongoDB in would require a larger abstraction of connection handling, triggers, and collectors rather than a simple option.

palazuelosagustin:
Adding JIRA to this PR for future reference.
https://perconadev.atlassian.net/browse/PT-2523

PR #1127, opened 2026-05-12
PT-2536 - Correct EMA calculation for WeightedAvgRate internal averages

marcelohpf:
Average number of rows and average time taken to process rows are convergent to the average by adding the 1-weight to the calculation.

Previous calculation was divergent and a not correct implementation of ema (Exponential Moving Average);

The result of update() is not affected due to the proportional calculation of number-rows/time-taken

You can verify the behavior of current calculation with :

```perl

use WeightedAvgRate;

my $weight = 0.8;
my $w = WeightedAvgRate->new( target_t => 10, weight => $weight );
my @fake_n = ( 10, 5, 3, 1, 10, 3, 5, 10, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20, 20 );

my ( $avg, $avg_ewma, $count, $sum ) = ( 0, 0, 0, 0 );
for my $n (@fake_n) {
    $sum   += $n;
    $count += 1;
    $avg = $sum / $count;
    $w->update( $n, 1 );
    $avg_ewma = $avg_ewma * $weight + $n * (1-$weight);
    printf( "%2.0f | %5.2f | %5.2f | %5.2f\n",
        $n, $avg, $w->{avg_n}, $avg, $avg_ewma );
}

1;
```

Root cause

The current calculation of average number of rows (avg_n) and average time (avg_t) seems to implement an smoothing algorithm to add some lag into the average convergence to protect estimation form sudden changes in a single datapoint, but also offer a faster convergence to that value on the long dataset (100+ data-points).

However, the current calculation is divergent serie because it does not apply a reduction factor to new data point every iteration.

As consequence, the avg_n and avg_t will increase forever and never converge to the "most common representative value";
Current formula is (avg * accumulative-factor) + new-value;
When it should be (avg * accumulative-factor) + (new-value * smooth-factor);

- [x] Documentation updated
- [x] Test suite update

### DIFF
--- a/bin/pt-online-schema-change
+++ b/bin/pt-online-schema-change
@@ -5659,8 +5659,8 @@ sub update {
    PTDEBUG && _d('Source op time:', $n, 'n /', $t, 's');
 
    if ( $self->{avg_n} && $self->{avg_t} ) {
-      $self->{avg_n}    = ($self->{avg_n} * $self->{weight}) + $n;
-      $self->{avg_t}    = ($self->{avg_t} * $self->{weight}) + $t;
+      $self->{avg_n}    = ($self->{avg_n} * $self->{weight}) + ($n * (1 - $self->{weight}));
+      $self->{avg_t}    = ($self->{avg_t} * $self->{weight}) + ($t * (1 - $self->{weight}));
       $self->{avg_rate} = $self->{avg_n}  / $self->{avg_t};
       PTDEBUG && _d('Weighted avg rate:', $self->{avg_rate}, 'n/s');
    }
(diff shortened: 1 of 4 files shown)


svetasmirnova:
Also tracked in Jira at https://perconadev.atlassian.net/browse/PT-2536

marcelohpf:
By re-reading the module name, I noticed that this was intended to be a variation of EWA with a specific weighted average for the proportion. Moving this into a simple EWA won't save any operation. This does same number of operations as current implementation, even if it did, 1 or 2 more calculations is nothing for modern computers.
```
avg_rate = avg_rate * weight + (1-weight) * $n/$t);
```

An weird implementation of EWA but the proportion of both values - copied rows / time to copy  - is what gives it a EWA behavior.

I also re-checked the convergency and as long 0 < weight <1, the series will converge. The term could be represented as:
f(value, weight) = value * [(1-weight^n)/(1-weight)], and that series converges to [value/(1-weight)] when n->infinity.

Even if not, it is limited to the number of rows present at the the table, which I doubt someone would have table with more rows than perl can handle with float-pointing

So I will be closing this PR.

PR #1140, opened 2026-06-22
Deprecation notice for pt-pg-summary

jobinau:
- [ x] The contributed code is licensed under GPL v2.0
- [ x] Contributor Licence Agreement (CLA) is signed
- [ x] Documentation updated

### DIFF
--- a/docs/percona-toolkit.pod
+++ b/docs/percona-toolkit.pod
@@ -120,7 +120,8 @@ ALTER tables without locking them.
 
 =item pt-pg-summary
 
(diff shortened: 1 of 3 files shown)


svetasmirnova:
We need a warning when somebody calls this tool.

jobinau:
> We need a warning when somebody calls this tool.

A clear warning is added. 
Since the screen might scroll very fast sometime,  The warning will be reappearing at the end of execution also.
DeprecationWarning

PR #1152, opened 2026-07-17
PMM-15195 Report the InnoDB redo log size on MySQL 8.0.30+ and 9.7 in pt-mysql-summary.

JiriCtvrtka:
[PMM-15195](https://perconadev.atlassian.net/browse/PMM-15195)

- [x] Documentation updated
- [x] Test suite update

PMM PR: https://github.com/percona/pmm/pull/5648
MySQL Exporter PR: https://github.com/percona/mysqld_exporter/pull/387
FB: https://github.com/Percona-Lab/pmm-submodules/pull/4479

## Summary
Updates `pt-mysql-summary` for MySQL 9.7 and removes an unnecessary historical
dependency on `mysqldump`.
- Reports InnoDB `Log File Size` from `innodb_redo_log_capacity` on MySQL
  8.0.30+, which is the value the server honours there even while the legacy
  variables are still present. `innodb_log_file_size` and
  `innodb_log_files_in_group` were deprecated in 8.0.30 and removed in 9.3.0 —
  measured: 9.0.1, 9.1.0 and 9.2.0 still expose all three variables, while
  9.3.0, 9.4.0 and 9.7.0 expose only the capacity.
- Retains the legacy `innodb_log_files_in_group * innodb_log_file_size`
  rendering for older MySQL versions.
- When only `innodb_log_file_size` is exposed (MariaDB 10.6+, which removed
  `innodb_log_files_in_group` and never had `innodb_redo_log_capacity`), reports
  that size as a bare total rather than discarding it as `Unknown`, matching how
  the redo capacity is rendered on 8.0.30+. MariaDB 10.5 still exposes the file
  count with value 1 — it was deprecated and ignored in 10.5.2 and only removed
  in 10.6.0 — so there the legacy `1 * 96.0M = 96.0M` rendering still applies.
  Measured on 10.5.29 and 10.6.28.
- Falls back to `Unknown` only when no redo size variable of any spelling is
  present. A capacity that is present but zero or non-numeric degrades to the
  legacy variables rather than being reported or turning into `Unknown`.
- Requires `mysqldump` only when schema analysis is explicitly requested with
  `--databases` or `--all-databases`; the default summary no longer fails when
  `mysqldump` is missing.
- Documents in `=head1 OUTPUT` that the `Log File Size` rendering is version
  dependent, naming the versions where each form applies, so the change from
  `count * per-file = total` to a bare total on 8.0.30+ is not a surprise.
- Bumps Go to 1.26.5 to address CVE-2026-39822 in the generated Go binaries and
  aligns `config/scripts/pt_builder.sh` with the version declared in `go.mod`.

## Compatibility
Default `pt-mysql-summary` and PMM Service Summary reports carry the same
information as before. On MySQL 8.0.30+ the `Log File Size` line prints the total
redo log capacity instead of `files * size = total`, because that capacity is
what the server honours from 8.0.30 on — and from 9.3.0 the per-file variables
are gone entirely. Schema analysis still requires `mysqldump`.

## Note for QA
This satisfies the ticket's Service Summary criterion differently than written.
The AC asked for `mysqldump` to be available in the execution PATH of the service
generating the report; this PR makes `mysqldump` optional for the default summary
instead, so no pmm-client packaging change is needed.

## Test plan
- Added coverage for the redo log variable layouts that actually differ, seven
  cases in total (`plan` 51 → 54): the 8.0.30 shape (all three variables
  present, capacity wins), the 9.3+ shape (capacity only), the MariaDB 10.6
  shape (`innodb_log_file_size` only), the MariaDB 10.5 shape (file count of
  one, which keeps the legacy rendering), the `Unknown` branch, and the two
  degradation paths where `innodb_redo_log_capacity` is present but zero or
  non-numeric.
- The last two matter because the guard is
  `[ "${redo_capacity}" -gt 0 ] 2>/dev/null`, which relies on `[` returning a
  non-zero status for a non-numeric operand with the error text suppressed.
  That is correct in bash, but it is the kind of idiom whose behaviour drifts
  between shells, so it is pinned by tests rather than by a comment.
- Verified by driving `section_innodb` directly against the `temp001` sample:
  all seven shapes produce the asserted output.
- The changed expected outputs in
  `t/pt-mysql-summary/samples/expected_output_temp_enc008.txt` and
  `..._enc009.txt` are a fix, not a test adjustment. Both samples carry all three
  variables (`innodb_log_file_size 5242880`, `innodb_log_files_in_group 2`,
  `innodb_redo_log_capacity 104857600`). On 8.0.30+ the legacy pair is deprecated
  and ignored, so the old `2 * 5.0M = 10.0M` reported a size the server does not
  honour; `100.0M` is the capacity actually in effect.
- Added tests asserting `mysqldump` is optional for the default summary and
  required for `--databases` / `--all-databases`.
- Verified the default summary and the reported redo log capacity against Oracle
  MySQL 9.7.
- Verified the PMM Client `mysqldump` binary can connect to MySQL 9.7.
- Built all Go binaries reported by Trivy using Go 1.26.5 and validated the
  package builder shell syntax.
- `bash -n` clean on `bin/pt-mysql-summary`, `lib/bash/report_mysql_info.sh` and
  `config/scripts/pt_builder.sh`.

## Changelog
Not updated here: the convention in this repo is release-time only — neither
`Changelog` nor `docs/release_notes.rst` has an unreleased section, and every
commit touching them is release preparation. Two user-visible items to pick up
at release prep:
- `Log File Size` output format change on MySQL 8.0.30+.
- `mysqldump` is no longer required for the default `pt-mysql-summary` run.

[PMM-15195]: https://perconadev.atlassian.net/browse/PMM-15195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


JiriCtvrtka:
@copilot review

JiriCtvrtka:
@copilot review

ademidoff (commented):
Two inline comments below. Two points that don't anchor to the diff:

### `=head1 OUTPUT` POD is now stale

`bin/pt-mysql-summary:3136` still documents:

    Log File Size | 2 * 5.0M = 10.0M

and the sample config just below at `bin/pt-mysql-summary:3245` reinforces it with `innodb_log_file_size = 5M`. Nothing tells the reader that on 8.0.30+ this field switches from `count * per-file = total` to a bare total. "Documentation updated - no changes needed" is fair for `=head1 SYSTEM REQUIREMENTS` (it never mentioned `mysqldump`), but the `OUTPUT` section now describes only the legacy rendering. A one-sentence note there should head off some "why did my summary format change?" reports.

### Changelog

`Changelog` isn't updated, and at least one user-visible item here arguably belongs in it: `Log File Size` output format change for MySQL 8.0.30+. Convention in this repo looks like release-time updates rather than per-PR, so if release prep picks these up, could you confirm both land in the next version? Otherwise a short entry in this PR would be good.

lib/bash/report_mysql_info.sh
```diff
@@ -1027,11 +1027,22 @@ section_innodb () {
...
+      name_val "Log File Size" "$(shorten ${redo_capacity} 1)"
+   else
+      local log_size="$(get_var innodb_log_file_size "$variables_file")"
+      local log_file="$(get_var innodb_log_files_in_group "$variables_file")"
+      if [ -n "${log_size}" ] && [ -n "${log_file}" ]; then
+         local log_total=$(awk "BEGIN {printf \"%.2f\n\", ${log_size}*${log_file}}" )
+         name_val "Log File Size" \
+                  "${log_file} * $(shorten ${log_size} 1) = $(shorten ${log_total} 1)"
+      else
+         name_val "Log File Size" "Unknown"
```
ademidoff, line 1048: Requiring **both** `log_size` and `log_file` means a server that exposes only `innodb_log_file_size` now reports `Unknown`, discarding a value it actually has. Verified by driving `section_innodb` with `innodb_log_files_in_group` stripped and no `innodb_redo_log_capacity` present:

    Log File Size | Unknown

That combination is MariaDB 10.5+, which dropped `innodb_log_files_in_group` and never had `innodb_redo_log_capacity`. The old code wasn't correct there either — it emitted a malformed ` * 92.0M = ` — so this isn't a clean regression, but it trades garbage for no information instead of for the number. Defaulting the multiplier costs one line:

```suggestion
      local log_size="$(get_var innodb_log_file_size "$variables_file")"
      local log_file="$(get_var innodb_log_files_in_group "$variables_file")"
      if [ -n "${log_size}" ]; then
         [ -n "${log_file}" ] || log_file=1
         local log_total=$(awk "BEGIN {printf \"%.2f\n\", ${log_size}*${log_file}}" )
         name_val "Log File Size" \
                  "${log_file} * $(shorten ${log_size} 1) = $(shorten ${log_total} 1)"
      else
         name_val "Log File Size" "Unknown"
```

Worth confirming whether MariaDB is in scope for `pt-mysql-summary` at all — there are no MariaDB samples under `t/pt-mysql-summary/samples/`, so CI wouldn't catch it either way.

t/lib/bash/report_mysql_info.sh
```diff
@@ -681,6 +681,24 @@ EOF
 
 test_format_innodb
 
+test_format_innodb_redo_capacity () {
+   local NAME_VAL_LEN=25
+   cp "$samples/temp001/mysql-variables" "$PT_TMPDIR/mysql-variables-redo-capacity"
+   printf 'innodb_redo_log_capacity\t104857600\n' >> "$PT_TMPDIR/mysql-variables-redo-capacity"
```
ademidoff: This appends `innodb_redo_log_capacity` to a sample that still contains **both** legacy variables, so the test proves *precedence* — the MySQL 8.0.30 case where all three exist. It doesn't cover the MySQL 9.7 shape (redo capacity present, legacy variables gone), which is what the ticket is actually about. The `Unknown` branch is untested too.

I confirmed the 9.7 path renders correctly by hand, but nothing in CI would catch a regression in it. One extra case closes the gap:

    grep -vE '^innodb_log_file_size|^innodb_log_files_in_group' \
       "$samples/temp001/mysql-variables" > "$PT_TMPDIR/mysql-variables-97"
    printf 'innodb_redo_log_capacity\t104857600\n' >> "$PT_TMPDIR/mysql-variables-97"

(with `plan 48` bumped accordingly)

JiriCtvrtka:
Hi @svetasmirnova, I saw that you opened this PR yesterday: https://github.com/percona/percona-toolkit/pull/1155. My ticket on the PMM side is to add support for MySQL 9.7. To implement it fully, I also created a PR for Percona Toolkit with the necessary changes. What do you think would be the best approach at this point?

Thank you.

lib/bash/report_mysql_info.sh
JiriCtvrtka: Fixed in e7bbf92, though not exactly as suggested — I added a separate branch rather than defaulting the multiplier, so MariaDB 10.5+ now prints 96.0M instead of 1 * 96.0M = 96.0M. The 1 * felt like noise, and a bare total matches how the redo capacity is rendered on 8.0.30+. Covered by the new "Format InnoDB log file size without a file count" case in t/lib/bash/report_mysql_info.sh, and documented in =head1 OUTPUT (f296151). Happy to switch to the 1 * X = X form if you'd rather keep the rendering uniform with the legacy path.

On MariaDB scope: I've kept the branch defensive rather than claiming support — there are still no MariaDB samples under t/pt-mysql-summary/samples/, so this only guarantees the variable layout is handled, not that the rest of the summary is validated against MariaDB. Adding real MariaDB fixtures feels out of scope for a MySQL 9.7 ticket, but I can file a follow-up if you think it's worth tracking.

4nte (approved):

JiriCtvrtka:
Hi @svetasmirnova could you check https://github.com/percona/percona-toolkit/pull/1152#issuecomment-5131378721? If we can decide about next steps.

bin/pt-mysql-summary
```diff
@@ -3157,6 +3169,15 @@ engine. The buffer pool fill percent and dirty percent are fuzzy-rounded. The
 last few lines are derived from the output of SHOW INNODB STATUS. It is likely
 that this output will change in the future to become more useful.
 
+The C<Log File Size> line above shows C<count * per-file = total>, which is how
+it is reported on servers exposing C<innodb_log_file_size> and
+C<innodb_log_files_in_group>. MySQL 8.0.30 replaced both with a single
+C<innodb_redo_log_capacity> and MySQL 9.0 removed them, so on those versions the
+line shows that capacity as a bare total instead. MariaDB 10.5 dropped
```
ademidoff: **This names the wrong MariaDB version, and it ships as `man pt-mysql-summary`.**

Measured on real servers:

```
mariadb:10.5.29  innodb_log_file_size 100663296   innodb_log_files_in_group 1   (no capacity)
mariadb:10.6.28  innodb_log_file_size 100663296   (no file count, no capacity)
```

So the bare-total branch applies from **10.6**, not 10.5. On 10.5 the file count is still there with value `1`, which lands in the legacy branch and renders `1 * 96.0M = 96.0M` — I confirmed that by running the function. A 10.5 user following this paragraph would expect `96.0M` and see something else.

The PR description has this right ("removed in 10.6.0", "MariaDB 10.5 still exposes the file count with value 1"), so it's the docs and comments that drifted, not the description. Same correction needed in `lib/bash/report_mysql_info.sh` and in the test comment in `t/lib/bash/report_mysql_info.sh`.

One line up, "MySQL 9.0 removed them" I could not confirm at 9.0 specifically: I measured 8.4.10 still exposing all three variables and 9.7.2 exposing only the capacity, so the removal happened somewhere in 9.x. 9.0 is plausible and matches the upstream changelog, but given the MariaDB version in the same sentence is wrong, it's worth a second look before this ships as documentation.

lib/bash/report_mysql_info.sh
```diff
@@ -1027,11 +1027,27 @@ section_innodb () {
...
+      name_val "Log File Size" "$(shorten ${redo_capacity} 1)"
+   else
+      local log_size="$(get_var innodb_log_file_size "$variables_file")"
+      local log_file="$(get_var innodb_log_files_in_group "$variables_file")"
+      if [ -n "${log_size}" ] && [ -n "${log_file}" ]; then
+         local log_total=$(awk "BEGIN {printf \"%.2f\n\", ${log_size}*${log_file}}" )
+         name_val "Log File Size" \
+                  "${log_file} * $(shorten ${log_size} 1) = $(shorten ${log_total} 1)"
+      elif [ -n "${log_size}" ]; then
+         # MariaDB 10.5 dropped innodb_log_files_in_group and never had
```
ademidoff: Same version correction as the POD block: this should be MariaDB **10.6**. 10.5.29 still reports `innodb_log_files_in_group` (value `1`), so it never reaches this branch — it renders `1 * 96.0M = 96.0M` through the legacy path above.

The branch itself is right, and reporting a size the server does expose rather than discarding it as `Unknown` is the correct call — it's only the version in the comment that's off.

t/lib/bash/report_mysql_info.sh
```diff
@@ -681,6 +681,68 @@ EOF
...
+   'innodb_redo_log_capacity\t104857600'
+
+# MySQL 9.x: the legacy variables were removed, only the capacity remains. This
+# is the layout the MySQL 9.7 work is actually about.
+test_format_innodb_log_size \
+   "Format InnoDB redo log capacity without legacy variables" "100.0M" \
+   '^innodb_log_file_size|^innodb_log_files_in_group|^innodb_redo_log_capacity' \
+   'innodb_redo_log_capacity\t104857600'
+
+# MariaDB 10.5+: innodb_log_files_in_group was dropped and the redo capacity
```
ademidoff: "MariaDB 10.5+" should be 10.6+ — see the POD comment. The value you chose is realistic, though: `100663296` is exactly what both 10.5.29 and 10.6.28 report for `innodb_log_file_size`.

Two gaps worth closing while the helper is fresh, since it makes each one a three-line call:

1. **The 10.5 shape has no test.** The four cases cover 8.0.30, 9.7, MariaDB-without-file-count and `Unknown`, but not the `file count = 1` shape that the PR description explicitly calls out as still applying. It works — I ran it and got `1 * 96.0M = 96.0M` — but it's the one documented rendering with nothing pinning it, and it's the one a future refactor of the `-n` checks would most easily break.

2. **The degradation paths aren't pinned.** `innodb_redo_log_capacity` present-but-zero and present-but-non-numeric both correctly fall through to the legacy branch, which I verified. `[ "${redo_capacity}" -gt 0 ] 2>/dev/null` relies on `[` returning a non-zero status for a non-numeric operand while the error text goes to `/dev/null` — correct in bash, but exactly the kind of idiom whose behaviour drifts between shells, and a comment won't catch that regression where a test would.

ademidoff (commented):
Reviewed as the third leg of PMM-15195, alongside percona/mysqld_exporter#387 and percona/pmm#5648. Rather than reading the branch, I drove the new `section_innodb` code directly against `lib/bash/report_mysql_info.sh` and checked every claim against real servers.

This is the cleanest of the three legs. The logic is correct on every shape I could produce, the test data reflects real servers, and the Go bump is legitimate. One real defect — the man page and two code comments name the wrong MariaDB version, and they contradict this PR's own description. Details inline.

## What I verified

Drove the new branch through eight shapes — your four plus four of my own:

| Shape | Output |
|---|---|
| 8.0.30 (all three vars) | `100.0M` |
| 9.7 (capacity only) | `100.0M` |
| MariaDB 10.6 (`log_file_size` only) | `96.0M` |
| MariaDB 10.5 (file count = 1) | `1 * 96.0M = 96.0M` |
| nothing present | `Unknown` |
| capacity = 0, no legacy | `Unknown` |
| capacity non-numeric, legacy present | falls back to legacy |
| capacity = 0, legacy present | falls back to legacy |

All eight correct — the `[ "${redo_capacity}" -gt 0 ] 2>/dev/null` guard degrades the way it needs to.

- **The changed expected outputs are a real fix, not a test fudge.** `t/pt-mysql-summary/samples/temp_enc008/mysql-variables` and `t/pt-mysql-summary/samples/temp_enc009/mysql-variables` carry all three variables (`innodb_log_file_size 5242880`, `innodb_log_files_in_group 2`, `innodb_redo_log_capacity 104857600`). On 8.0.30+ the legacy pair is deprecated and ignored, so the old `2 * 5.0M = 10.0M` was actively wrong and `100.0M` is the capacity the server actually honours. Worth stating that explicitly in the description — as written, a reviewer could read those two sample changes as adjusting tests to match new behaviour.
- **The `check_mysql` gating is safe.** `OPT_ALL_DATABASES` is a real option and the new guard matches the existing idiom at `bin/pt-mysql-summary:1177` and `bin/pt-mysql-summary:2564`. `CMD_MYSQLDUMP` is referenced only at `bin/pt-mysql-summary:1081` and `bin/pt-mysql-summary:1091`, both inside the schema-dump path gated on those same options, so no default-path use survives and the check can safely be conditional.
- **The `bin/` vs `lib/` comment divergence is expected**, not hand-editing — `util/update-modules:92` pipes module content through `grep -v '^ *#'`. I checked because the two blocks differ; they differ correctly.
- `go build ./...` clean under Go 1.26.6, and `bash -n` clean on `bin/pt-mysql-summary`, `lib/bash/report_mysql_info.sh` and `config/scripts/pt_builder.sh`.
- **CVE-2026-39822 checks out**: `os.Root` improperly follows symlinks outside the root on Unix when the final path component is a symlink and the path ends in `/` (CWE-61, CVSS 7.8), fixed in Go 1.25.12 / 1.26.5 / 1.27.0-rc.2. Both the bump and the citation are accurate.

## Note for QA

This leg satisfies the ticket's Service Summary criterion differently than written. The AC asked for `mysqldump` to be available in the execution PATH of the service generating the report; this PR makes it optional instead, which is the better fix. Worth flagging on the ticket so QA doesn't go looking for a pmm-client packaging change that no longer needs to happen.

bin/pt-mysql-summary
JiriCtvrtka: Fixed — and you were right to doubt the other half of that sentence too, it isn't 9.0 either.

Measured across the 9.x line: 9.0.1, 9.1.0 and 9.2.0 all still expose innodb_log_file_size and innodb_log_files_in_group alongside the capacity; 9.3.0, 9.4.0 and 9.7.0 expose only the capacity. So the removal was 9.3.0. And on MariaDB, 10.5.29 reports innodb_log_files_in_group = 1 while 10.6.28 doesn't, so the bare-total branch starts at 10.6 exactly as you said.

The POD now says 9.3 and 10.6, plus a sentence that 10.5 keeps the count * per-file = total form. I've also corrected the description bullet that implied the legacy variables were already gone on 8.0.30+ — they're deprecated but present there, which is the whole reason the capacity has to win.

lib/bash/report_mysql_info.sh
JiriCtvrtka: Corrected to 10.6, with a follow-up line noting that 10.5 still reports the file count as 1 and therefore takes the legacy branch above and renders 1 * 96.0M = 96.0M. Verified on 10.5.29 and 10.6.28.

t/lib/bash/report_mysql_info.sh
JiriCtvrtka: Comment corrected to 10.6+, and both gaps closed while the helper was fresh — plan 51 → 54:

"Format InnoDB log file size with a file count of one" — the 10.5 shape, pinning the 1 * 96.0M = 96.0M rendering you measured.
"Format InnoDB log file size when the capacity is zero" and "...when the capacity is not a number" — both degradation paths through [ "${redo_capacity}" -gt 0 ] 2>/dev/null, so the shell-behaviour dependency you flagged is pinned by tests rather than by a comment.
Verified the same way you did, driving section_innodb against the temp001 sample: all seven shapes produce the asserted output, including your four extras.

JiriCtvrtka:
> Reviewed as the third leg of PMM-15195, alongside [percona/mysqld_exporter#387](https://github.com/percona/mysqld_exporter/pull/387) and [percona/pmm#5648](https://github.com/percona/pmm/pull/5648). Rather than reading the branch, I drove the new `section_innodb` code directly against `lib/bash/report_mysql_info.sh` and checked every claim against real servers.
> 
> This is the cleanest of the three legs. The logic is correct on every shape I could produce, the test data reflects real servers, and the Go bump is legitimate. One real defect — the man page and two code comments name the wrong MariaDB version, and they contradict this PR's own description. Details inline.
> 
> ## What I verified
> Drove the new branch through eight shapes — your four plus four of my own:
> 
> Shape	Output
> 8.0.30 (all three vars)	`100.0M`
> 9.7 (capacity only)	`100.0M`
> MariaDB 10.6 (`log_file_size` only)	`96.0M`
> MariaDB 10.5 (file count = 1)	`1 * 96.0M = 96.0M`
> nothing present	`Unknown`
> capacity = 0, no legacy	`Unknown`
> capacity non-numeric, legacy present	falls back to legacy
> capacity = 0, legacy present	falls back to legacy
> All eight correct — the `[ "${redo_capacity}" -gt 0 ] 2>/dev/null` guard degrades the way it needs to.
> 
> * **The changed expected outputs are a real fix, not a test fudge.** `t/pt-mysql-summary/samples/temp_enc008/mysql-variables` and `t/pt-mysql-summary/samples/temp_enc009/mysql-variables` carry all three variables (`innodb_log_file_size 5242880`, `innodb_log_files_in_group 2`, `innodb_redo_log_capacity 104857600`). On 8.0.30+ the legacy pair is deprecated and ignored, so the old `2 * 5.0M = 10.0M` was actively wrong and `100.0M` is the capacity the server actually honours. Worth stating that explicitly in the description — as written, a reviewer could read those two sample changes as adjusting tests to match new behaviour.
> * **The `check_mysql` gating is safe.** `OPT_ALL_DATABASES` is a real option and the new guard matches the existing idiom at `bin/pt-mysql-summary:1177` and `bin/pt-mysql-summary:2564`. `CMD_MYSQLDUMP` is referenced only at `bin/pt-mysql-summary:1081` and `bin/pt-mysql-summary:1091`, both inside the schema-dump path gated on those same options, so no default-path use survives and the check can safely be conditional.
> * **The `bin/` vs `lib/` comment divergence is expected**, not hand-editing — `util/update-modules:92` pipes module content through `grep -v '^ *#'`. I checked because the two blocks differ; they differ correctly.
> * `go build ./...` clean under Go 1.26.6, and `bash -n` clean on `bin/pt-mysql-summary`, `lib/bash/report_mysql_info.sh` and `config/scripts/pt_builder.sh`.
> * **[CVE-2026-39822](https://github.com/advisories/GHSA-xcgv-8mv7-v8c7) checks out**: `os.Root` improperly follows symlinks outside the root on Unix when the final path component is a symlink and the path ends in `/` (CWE-61, CVSS 7.8), fixed in Go 1.25.12 / 1.26.5 / 1.27.0-rc.2. Both the bump and the citation are accurate.
> 
> ## Note for QA
> This leg satisfies the ticket's Service Summary criterion differently than written. The AC asked for `mysqldump` to be available in the execution PATH of the service generating the report; this PR makes it optional instead, which is the better fix. Worth flagging on the ticket so QA doesn't go looking for a pmm-client packaging change that no longer needs to happen.

Thanks — all three inline items are fixed, and one of them turned out to be two bugs rather than one: the MySQL removal version was wrong as well as the MariaDB one. It's 9.3.0, measured across 9.0.1 / 9.1.0 / 9.2.0 (all three variables present) versus 9.3.0 / 9.4.0 / 9.7.0 (capacity only). Details in the threads.

On the Go bump being unrelated scope — I'd push back on that one, because it turns out to be what the build gate requires rather than optional hardening. That job ends in a Trivy scan with severity: CRITICAL,HIGH, ignore-unfixed: true, exit-code: 1. I reproduced it locally against the same image (oraclelinux:9-slim + microdnf -y update + COPY bin/*) and got exit 1 with 10 HIGH findings, all Go and no OS:

stdlib v1.26.5 — 8 HIGH, fixed in 1.26.6: encoding/asn1, x/net/idna, x/net/dns/dnsmessage, net/http, html/template, encoding/xml, net/url, crypto/tls
golang.org/x/net v0.55.0 — CVE-2026-46600, fixed in 0.56.0
golang.org/x/text v0.38.0 — CVE-2026-56852, fixed in 0.39.0
So the 1.26.5 pin was correct when written — it fixed CVE-2026-39822 — but 1.26.6 added those eight and the pin went stale against a database that moves independently of the branch. That also explains why it built clean for you: you used 1.26.6. Bumping to 1.26.7 plus x/net 0.56.0 and x/text 0.39.0 and re-running the identical scan gives exit 0, no findings. Description updated accordingly.

On the Changelog: confirmed, the convention here is release-time only — neither Changelog nor docs/release_notes.rst has an unreleased section, and every commit touching them is release prep. The two items to pick up at release prep are the Log File Size output change on 8.0.30+ and mysqldump becoming optional for the default summary. Both are now called out in the description.

On the QA note: agreed and added to the description, and I'll put it on the ticket so nobody chases a pmm-client packaging change.

Still outstanding and not mine to settle: #1155 touches the same section_innodb hunk with incompatible behaviour, and I haven't had a reply from @svetasmirnova since 30 Jul or 20 Aug. That's the real blocker on this leg.

ademidoff (approved):

svetasmirnova:
@JiriCtvrtka https://github.com/percona/percona-toolkit/pull/1155 is not only about pt-mysql-sumary, it is about if the toolkit can work with 9.7. It is merged now, and I am reviewing this PR. pt-mysql-summary requires more changes, see https://perconadev.atlassian.net/browse/PT-2505 So this will be the first step in improving pt-mysql-summary work.

svetasmirnova (approved):

JiriCtvrtka:
@svetasmirnova Thank you for merging this one.

PR #1155, opened 2026-07-29
PT-2548 - Mysql 9.7 support

svetasmirnova:
- [ ] Documentation updated
- [ ] Test suite update


lib/bash/report_mysql_info.sh
```diff
@@ -284,7 +284,7 @@ format_status_variables () {
 }
 
 # Slices the processlist a bunch of different ways.  The processlist should be
-# created with the \G flag so it's vertical.
+# created in vertical formatinnodb_log_file_size.
```
svetasmirnova:
```suggestion
# created in vertical format.
```

lib/OobNibbleIterator.pm
```diff
@@ -80,7 +80,7 @@ sub new {
       PTDEBUG && _d('Past lower statement:', $past_lower_sql);
 
       my $explain_past_lower_sql
-         = "EXPLAIN SELECT "
+         = "EXPLAIN /*! FORMAT=TRADITIONAL */ SELECT "
```
svetasmirnova:
```suggestion
         = "EXPLAIN /*!90700 FORMAT=TRADITIONAL */ SELECT "
```

lib/OobNibbleIterator.pm
```diff
@@ -97,7 +97,7 @@ sub new {
       PTDEBUG && _d('Past upper statement:', $past_upper_sql);
 
       my $explain_past_upper_sql
-         = "EXPLAIN SELECT "
+         = "EXPLAIN /*! FORMAT=TRADITIONAL */ SELECT "
```
svetasmirnova:
```suggestion
         = "EXPLAIN /*!90700 FORMAT=TRADITIONAL */ SELECT "
```

t/pt-pmp/pt-pmp.t
```diff
@@ -15,6 +15,11 @@ use Test::More;
 
 my $sample = "$trunk/t/pt-pmp/samples";
 
+# We need to perform this check, because uutils is default in Ubuntu 26
+if ( $? >> 8 == 0 ) {
```
EvgeniyPatlan, line 19: No command runs before this check, so $? is 0
Maybe we should add something like this before if:
`sort --version 2>/dev/null | grep -q uutils`;

lib/RowChecksum.pm
```diff
@@ -289,7 +292,7 @@ sub _get_hash_func {
    }
    my ($dbh) = @args{@required_args};
    my $o     = $self->{OptionParser};
-   my @funcs = qw(CRC32 FNV1A_64 FNV_64 MURMUR_HASH MD5 SHA1);
+   my @funcs = qw(CRC32 FNV1A_64 FNV_64 MURMUR_HASH MD5 SHA1 SHA2);
```
EvgeniyPatlan, line 295: Adding SHA2 here (and at lib/TableChecksum.pm:158) makes --function=SHA2 reachable, but SHA2(x,256) returns 64 chars and _get_crc_width returns 64 — while the replicate table created from MAGIC_create_replicate (bin/pt-table-checksum, create_repl_table) is:
this_crc       CHAR(40)     NOT NULL,
source_crc     CHAR(40)         NULL,
With strict sql_mode that's ER_DATA_TOO_LONG on every chunk; without strict mode it truncates silently. SHA1 fits at exactly 40, so SHA2 is the first function to exceed it. So we need to extend this_crc and source_crc to CHAR(64)

lib/bash/report_mysql_info.sh
```diff
@@ -1029,9 +1029,15 @@ section_innodb () {
...
-            "${log_file} * $(shorten ${log_size} 1) = $(shorten ${log_total} 1)"
+   if [ -n "$log_size" -a -n "$log_file" ]; then
+      local log_total=$(awk "BEGIN {printf \"%.2f\n\", ${log_size}*${log_file}}" )
+      name_val "Log File Size"       \
+               "${log_file} * $(shorten ${log_size} 1) = $(shorten ${log_total} 1)"
+   else
+      local log_total="$(get_var innodb_redo_log_capacity "$variables_file")"
+      name_val "Redo Log Capacity"       \
+               "$(shorten $(get_var innodb_redo_log_capacity "$variables_file") 0)"
+   fi
```
EvgeniyPatlan, line 1040: log_total is assigned but never read, and get_var is called twice for the same value. maybe we need to do something like that:
```
local log_total="$(get_var innodb_redo_log_capacity "$variables_file")"
name_val "Redo Log Capacity" "$(shorten ${log_total} 0)"
```

lib/bash/report_mysql_info.sh
```diff
@@ -1029,9 +1029,15 @@ section_innodb () {
 
    local log_size="$(get_var innodb_log_file_size "$variables_file")"
    local log_file="$(get_var innodb_log_files_in_group "$variables_file")"
-   local log_total=$(awk "BEGIN {printf \"%.2f\n\", ${log_size}*${log_file}}" )
-   name_val "Log File Size"       \
-            "${log_file} * $(shorten ${log_size} 1) = $(shorten ${log_total} 1)"
+   if [ -n "$log_size" -a -n "$log_file" ]; then
```
EvgeniyPatlan, line 1032: -a is outdated.
better: 
`[ -n "$log_size" ] && [ -n "$log_file" ]`

lib/RowChecksum.pm
svetasmirnova: Fixed

lib/bash/report_mysql_info.sh
svetasmirnova: Fixed

lib/bash/report_mysql_info.sh
svetasmirnova: Fixed

bin/pt-stalk
```diff
@@ -1577,7 +1576,9 @@ purge_samples() {
    if [ $targetSize -gt 0 ]; then
       files_to_delete=$(find "$dir" -maxdepth 1 -type f -exec basename {} \; | cut -f1 -d- | sort -r | uniq | tail -n +1)
       for prefix in $files_to_delete; do
-          current_size=$(du -BM $dir | cut -f1 -d"M")
+          # We need to use chain of cut -f1 | cut -f1 -d"M" calls due to bug in du from uutils
+          # https://github.com/uutils/coreutils/issues/13605
+          current_size=$(du -BM $dir | cut -f1 | cut -f1 -d"M")
```
EvgeniyPatlan, line 1581:
```suggestion
current_size=$(du -s -BM "$dir" | awk '{gsub(/M/, "", $1); print $1}')
```

lib/TableChunker.pm
```diff
@@ -589,11 +592,22 @@ sub _chunk_char {
...
+      if ( !$old_autocommit ) {
+         my $gtid_mode = eval { $dbh->selectrow_array('SELECT @@GLOBAL.gtid_mode') };
+         $need_autocommit = !$EVAL_ERROR && defined $gtid_mode && uc($gtid_mode) ne 'OFF';
+      }
+      $dbh->{AutoCommit} = 1 if $need_autocommit;
       my $col_def = $args{tbl_struct}->{defs}->{$chunk_col};
       $sql        = "CREATE TEMPORARY TABLE $tmp_db_tbl ($col_def) "
                   . "ENGINE=MEMORY DEFAULT CHARSET = utf8";
       PTDEBUG && _d($dbh, $sql);
       $dbh->do($sql);
```
EvgeniyPatlan, line 609: No eval around the CREATE TEMPORARY TABLE. If we 'die' here and exit - autocommit will be 1

bin/pt-table-checksum
```diff
@@ -13929,9 +13931,9 @@ structure (MAGIC_create_replicate):
      chunk_index    VARCHAR(200)     NULL,
      lower_boundary TEXT             NULL,
      upper_boundary TEXT             NULL,
-     this_crc       CHAR(40)     NOT NULL,
+     this_crc       CHAR(64)     NOT NULL,
```
EvgeniyPatlan, line 13934: create_replicate_table() uses CREATE TABLE IF NOT EXISTS. So if old table was present it will be truncated to CHAR(40) .

lib/RowChecksum.pm
```diff
@@ -298,7 +301,7 @@ sub _get_hash_func {
    my $error;
    foreach my $func ( @funcs ) {
       eval {
-         my $sql = "SELECT $func('test-string')";
+         my $sql = "SELECT $func('test-string'" . ( uc $func eq 'SHA2' ? ", $sha2_length)" : ")");
```
EvgeniyPatlan, line 304: SHA2() returns NULL when the server was built without SSL support. So we will pick SHA2 and will compare every row with NULL

lib/TableChecksum.pm
```diff
@@ -152,16 +155,17 @@ sub is_hash_algorithm {
...
       unshift @funcs, $args{function};
    }
    my ($result, $error);
    do {
       my $func;
+      my $sql;
       eval {
          $func = shift(@funcs);
-         my $sql = "SELECT $func('test-string')";
+         $sql = "SELECT $func('test-string'" . ( uc $func eq 'SHA2' ? ", $sha2_length)" : ")" );
```
EvgeniyPatlan, line 168: SHA2() returns NULL when the server was built without SSL support. So we will pick SHA2 and will compare every row with NULL

EvgeniyPatlan (changes requested):

t/lib/MasterSlave.t
```diff
@@ -847,7 +847,7 @@ like(
 
 SKIP: {
 
-   skip "Only test on mysql 5.7",6 if ( $sandbox_version lt '5.7' );
+   skip "Only test on mysql 5.7 or newer", if ( $sandbox_version lt '5.7' );
```
EvgeniyPatlan:
```suggestion
   skip "Only test on mysql 5.7 or newer" if ( $sandbox_version lt '5.7' );
```

It works only because this file uses done_testing.

bin/pt-stalk
svetasmirnova: du -s returns single line, I need list.

E.g., 

```
sveta@lemurr:~/mysqlpackages/mariadb-12.3.2-linux-systemd-x86_64/mariadb-test$ du -BM .
61	./main
5	./include
4	./var/install.db/mysql
1	./var/install.db/test
1	./var/install.db/performance_schema
1	./var/install.db/sys
1	./var/install.db/mtr
57	./var/install.db
...
sveta@lemurr:~/mysqlpackages/mariadb-12.3.2-linux-systemd-x86_64/mariadb-test$ du -s -BM .
421	.
```

lib/TableChunker.pm
svetasmirnova: We are setting session-level autocommit. If we die, the session will be closed, and the tool will stop. So the value of the autocommit does not matter anymore.

bin/pt-table-checksum
svetasmirnova: This is fine. The default hash function is CRC32.

lib/RowChecksum.pm
svetasmirnova: It is not possible to compile MySQL without SSL support in supported versions, so I would not add this exception.

lib/TableChecksum.pm
svetasmirnova: It is not possible to compile MySQL without SSL support in supported versions, so I would not add this exception.

hrvojem:
LGTM

hrvojem (approved):

EvgeniyPatlan (approved):
LGTM

PR #1171, opened 2026-09-07
Pt 2531 pt online schema change creates sleeping threads in replicas

svetasmirnova:
- [ ] Documentation updated
- [ ] Test suite update


saikumar-vs (approved):

lib/MasterSlave.pm
```diff
@@ -1120,6 +1132,20 @@ sub get_cxn_from_dsn_table {
...
+      foreach my $new_replica ( @cxn ) {
+         if ( $known_replica->{dsn}->{h} eq $new_replica->{dsn}->{h}
+               and $known_replica->{dsn}->{P} eq $new_replica->{dsn}->{P}
+               and $known_replica->{dsn}->{u} eq $new_replica->{dsn}->{u}
+               and $known_replica->{dsn}->{p} eq $new_replica->{dsn}->{p}
+         ) {
+            next REPLICAS;
+         }
+         $known_replica->{dbh}->disconnect;
+      }
```
matias-sanchez:
```suggestion
      }
      $known_replica->{dbh}->disconnect if $known_replica->{dbh};
```

after test i saw this one decides to disconnect before it finishes searching: if the first entry of the list doesn't match, it disconnects right there without looking at the remaining ones. So a replica that is still in the dsns table, just not in the first position, gets disconnected anyway.

Moving it after the inner loop makes it run once per replica. The if avoids a die when a replica has no dbh.

lib/MasterSlave.pm
matias-sanchez:
```suggestion
   my $dsn_tbl_cxn = $self->{dsn_tbl_cxn}->{$dsn_table_dsn} ||= $make_cxn->(dsn => $dsn);
```

i saw that this one opens a new connection to the dsns table host on every refresh. Reusing it as this removed the reconnection

matias-sanchez (commented):
adding enhancements based on testing with 1 source and 4 replicas in the dsns table. With both applied i saw no reconnections over many chunks.

EvgeniyPatlan (approved):

